PR #748, opened 2014-08-13
ToolBarMenu represents menu of view. It should be renamed

rhopp:
Class ToolBarMenu represents menu of View. Thus it should be renamed to something like ViewMenu. 
Also lookup logic contains some errors and does not have any checks whether menu was found or not.


rhopp:
I'm closing this issue and creating new one.

PR #896, opened 2014-12-01
Capture screenshot when requirement fails

mlabuda:


mlabuda:
Link to execution parameter documentation https://github.com/jboss-reddeer/reddeer/wiki/RedDeer-execution-parameters

mlabuda:
test this

mlabuda:
test this

vpakan:
test this

vpakan:
* https://github.com/jboss-reddeer/reddeer/pull/896/files#diff-60fb091d04260ced11054120bfaba9a0R44 - equalsIgnoreCase() method should be used
* https://github.com/jboss-reddeer/reddeer/pull/896/files#diff-60fb091d04260ced11054120bfaba9a0R57 - System.clearProperty("captureScreenshot") should be used to have the same environment as was before

vpakan:
System.clearProperty("captureScreenshot") should be used in case related property was not defined before it was changed by test in case this property was set to true before it was changed by test it should be set back to true when test is finished

PR #910, opened 2014-12-16
dev_Notify Tree when item is selected

rawagner:


rawagner:
Seems like this is no longer an issue. Closing

PR #906, opened 2015-02-05
Initial code generator

apodhrad:
Add a feature to generate code from wizards, shells, views, etc.

How will it work?
Just focus on a wizard or view a press a key shortcut with will generate a RedDeer code into a view or some standalone shell.


apodhrad:
In the future the recorder will depends on these generators

vpakan:
* CodeGeneratorViewTest.openCodeGeneratorViewTest default
* org.jboss.reddeer.generator.test about.html should contain year 2015 instead of 2010
* org.jboss.reddeer.generator should also contain about.html probably
* can you use different shortcut than CTRL+SHIFT for starting generator this shortcut is used via Eclipse Spy

apodhrad:
* Now the shortcut is Alt + g
* Feature is not added to site/category.xml since it is part of recorder feature (in the future recorder will dependent on generator bundle)
* The rest should be fixed

vpakan:
plugins/org.jboss.reddeer.generator/build.properties - should contain about.html

apodhrad:
done

PR #948, opened 2015-02-24
BrowserEditor can be created with any editor.

rhopp:
This will fail, because not every editor has InternalBrowser in it.
Solution is to ensure, that the editor is org.eclipse.ui.internal.browser.WebBrowserEditor


rhopp:
test this

rhopp:
retest this

rawagner:
Is there any reason to have constrain on org.eclipse.ui.browser version ?

rhopp:
Nope... It just slipped through my fingers. Fixed.

rawagner:
PTM

PR #955, opened 2015-02-26
Cannot find active view on Eclipse Mars

ljelinkova:
On Mars, the method IWorkbenchPage.getViewReferences()) does not return views that are open but not on the active tab. We can use IWorkbenchPage.getViewStack(view) to access them.


rawagner:
PTM

PR #868, opened 2015-03-03
Enhance problem view

mlabuda:
- getting problems
- getting error
- filtering
- filter method consider fixed order of columns, this is bad assumption


mlabuda:
test this

vpakan:
Mark old nethod getWarnings() and getErrors() as deprecated. They should directly call new methods with respective names and 2 new parameters

PR #900, opened 2015-03-04
Handle preference page changed/dialog ok

rawagner:
OK or changing page might call some another shell asking for confirmation. We need to handle this - add hook to ok dialog method and listener for tree selection - page should implement what needs to be done to handle another dialog. See also eclipse implementation to see how its handled there (if page has two different methods or just one for both events).


rawagner:
Depends on #963

rawagner:
test this

rawagner:
test this

rawagner:
test this

rawagner:
test this

rawagner:
Doc issue created #1059

rawagner:
test this

rawagner:
retest this

rawagner:
retest this

rawagner:
keep track of all previously opened pages
page should know shell title, or null if there's none

dialog should check all previously opened pages, check shell title and do the handling

rawagner:
retest this

rawagner:
retest this

PR #951, opened 2015-03-05
Cannot catch EclipseLayerException with new RedDeerException

mlabuda:
After the RedDeerException has been moved to common and the old one in swt plugin deprecated, you actually cannot use the new one to catch EclipseLayerException since it is not its descendant.


ljelinkova:
Change all exceptionw (SWTLayer, Eclipse, etc) to extend `org.jboss.reddeer.swt.exception.RedDeerException` so that they can be caught by both, `org.jboss.reddeer.swt.exception.RedDeerException` and `org.jboss.reddeer.common.exception.RedDeerException` exception.

vpakan:
test this

vpakan:
So now we have 2 RedDeerException witihin 2 different plugins/packages it doesn't seems right. What about to rename the one from swt plugin to SWTRedDeerException to make ti cleat when to use it.

If yes we should check also documentation and update it where appropriate.

ljelinkova:
This is just a temporary state where `org.jboss.reddeer.swt.exception.RedDeerException`is deprecated and will be removed in 0.8.0. This fix should allow user to catch all exceptions with both, `org.jboss.reddeer.swt.exception.RedDeerException` and `org.jboss.reddeer.common.exception.RedDeerException` so that his tests are not broken.

PR #909, opened 2015-03-05
Requirements should cleanup after themselves

rawagner:


rawagner:
Be aware that this will break jbosstools-integration-tests reddeer plugins because I added a method to Requirement interface

ljelinkova:
- please add comments to public classes and methods
- I think that requirements should be cleaned up after afterClass methods
```
    @Override
    public void evaluate() throws Throwable {
        requirements.cleanUp();
        statement.evaluate();
    }
```
- there is test suite `RequirementsTestSequenceSuite` testing the proper order of requirements fullfilling, @Before, @BeforeClass etc. methods, it should be extended also for cleanup method
- in `ServerRequirement.cleanup()` I do not think you should have `lastServerConfiguration != null` in the condition
- you should extend tests for `ServerRequirement` to test the cleanup behavior

rawagner:
test this

rawagner:
I've fixed/added all things as you suggested but i kept  `lastServerConfiguration != null`

rawagner:
test this

ljelinkova:
test this

ljelinkova:
Pushed to master

ljelinkova:
@rawagner  Please create Jira for doc update

rawagner:
created #1056

PR #779, opened 2015-03-05
dev_ClassMatcher should allow to match also specific class

mlabuda:


ljelinkova:
This may effect many things, moving to 0.8.0

PR #962, opened 2015-03-10
Problems view - allow to filter on various number of columns by matchers

mlabuda:


vpakan:
* getErrorProblems() and getWarningProblems() methods should keep their old names getErrors() and getWarnings() but they should change their return type to new class Problem with fields equals problems table columns i. e. description, location ... and method boolean isError()
* change test to use these methods and use at least one problem matcher with String input instead of Matcher input in constructor. Right now test is using fixed column index in assert methods what could become a problem when order of column will be changed or some column will be added

vpakan:
ProblemsExists.java should be refactored:
* deprecated constructor should call non deprecated one
* test method should use only non deprecated methods and enum and should contain something like this return !ProblemsView.getProblems (problemType).empty() so we will loose support for ProblemType.BOTH but it's fine for me it's actually not existing problem type in Problems view. User always can wait for warning first and error later or vice versa. I'm not sure if we should implement many ProblemType(s) as parameter for condition constructor

vpakan:
* add javadoc to public classes methods and constructors

PR #968, opened 2015-03-13
SyncExec before and after interceptor support for RedDeer

jpeterka:
This should provide facility to inject selected user registration operations (like
 logging, diagnostic, etc.) around  Display.syncExec() It's not expected that user provides UI operation that changes UI state. If so user is responsible  for restoring original state.


vpakan:
* is IBeforeSync.java needed/used?
* SyncInterceptorManager.isIntercepted() has mistyped javadoc
* I think  usage of SyncInterceptorManager.isIntercepted() is not needed it will never get nested call. I returns true only when interceptor methods are called and they will be called again until it's finished first

jpeterka:
1) ok - remoed, my mistake
2) javadoc fixed
3) it protect for cycling interceptor calls inside sync exec when interceptor method utilizes syncExec
4) later, ok?
5) I don't know why, it's disabled by default and enabling/disabling is completely up to user by it's usage

jpeterka:
Tests added

vpakan:
5) Imagine you want to have interceptor available only in case something went wrong when running tests. You have to have then in code and registered but would be nice to enable/disable them from command line something like adding more debugging but only in case tests are failing

jpeterka:
First I would not add another switch until there is real demand for it. Also I don't like idea to enable or disable it in bulk. Test developer should decide if he wants to use it or not and has complete freedom to implement command line switch for each individual Interceptor or for group of them which gives him finer tuning. Disabling/enabling is on level or registering and user decides when to register each Interceptor. Note that idea of our extensions (like reddeer.close.shells). They are also controlled individually and not by some total extension switch. For your use case. If something goes wrong when running test (even on Jenkins) and if you have some diagnostic Interceptor I would suggest to implement switch for your particular Interceptor so you can enable/disable it as you need.

jpeterka:
There si some doc draft https://github.com/jboss-reddeer/reddeer/wiki/Handling-around-syncExec-with-Interceptors

vpakan:
On one side you enable intercepting everywhere in code where syncExec() is run even if in most cases it's not relevant but you don't want to have switch to disable it completely to at least be able to prove that interceptors are reason for breaking build if such situation will happen. Never mind it was just idea it's not mandatory to me to push this PR.
I have demand for it now. Implement it when it will be needed could be annoying. I see no problem to have one more system property used. It's up to you.
Don't forget to create issue for documenting this feature please.

jpeterka:
Doc issues: https://github.com/jboss-reddeer/reddeer/issues/972

PR #993, opened 2015-03-24
It's not possible to wait for particular shell to disappear.

rhopp:
Extend `ShellIsActive` condition to be able to use this condition against particular shell.


jpeterka:
Better to wait until #996 is pushed and apply it into core

PR #996, opened 2015-03-25
Initial swt parts refactoring into core and common

jpeterka:


jpeterka:
Usage re-factoring will be in next PR to keep PR "smaller"

PR #1015, opened 2015-04-08
Deprecate wizard getPage

rawagner:
user will construct pages by himself


jpeterka:
Please remove usage of these deprecated methods from rest of RedDeer

PR #1027, opened 2015-04-08
Remove singleton atribute from org.jboss.reddeer.junit

mmalina:
The reason Vlado defined org.jboss.reddeer.junit is to save
some memory in osgi. But this poses a problem when we then want
to use two or more different RedDeer versions as a dependency on the same
update site - tycho will not allow to build a site that relies on different versions
of a bundle that is defined as a singleton. So we concluded that this is more
important.


mmalina:
I accidentally created this topic branch on the project directly. I will close this PR and open a new one.

PR #1029, opened 2015-04-08
Remove singleton atribute from RedDeer plugins

mmalina:
The reason we defined some plugins as singletons is to save
some memory in osgi. But this poses a problem when we then want
to use two or more different RedDeer versions as a dependency on the same
update site - tycho will not allow to build a site that relies on different versions
of a bundle that is defined as a singleton. So we concluded that this is more
important.


mmalina:
test this

mmalina:
It turns out that bundles that use or have an extension point (or something like that) have to be singletons, so this breaks things and we have to reject this. (There were around 150 test failures with this applied.)

PR #1031, opened 2015-04-09
Reddeer.jface is indirectly referenced by eclipse plugin

jpeterka:
This code:
```java
		NewJavaProjectWizardDialog pw = new NewJavaProjectWizardDialog();
		pw.open();
```
results in error when reddeer jface plugin in not added as a dependency because of indirect reference:
Multiple markers at this line
	- The type org.jboss.reddeer.jface.wizard.NewWizardDialog cannot be resolved. It is 
	 indirectly referenced from required .class files
	- The method open() is undefined for the type NewJavaProjectWizardDialog

jface should be reexported in reddeer.eclipse plugin


jpeterka:
Verification job looks good, moving to gk

jpeterka:
Blue job after after commit update

PR #1032, opened 2015-04-14
Deprecate swt's Regex and RegexSeq

rawagner:
It's not used anywhere in reddeer. User can use Matcher


rawagner:
I killed the job, because there's really no need to run it.

PR #1041, opened 2015-04-16
Dont save editors when closing

rawagner:
Currently when calling WorkbenchParHandler.closeAllEditors() we send true to activePage.closeEditors(). This however means, according to eclipse javadoc, that user will be given chance to save editors (which means shell asking for confirmation will popup) - this might result in locking this operation. Safer would be to send false.


rawagner:
test this

rawagner:
test this

rawagner:
test this

rawagner:
test this

vpakan:
Remove method closeAllEditors() from WorbenchLookup and use similar method from EditorLookup instead

PR #1042, opened 2015-04-20
Wait for shell active in quickfix wizard

rawagner:


apodhrad:
looking

apodhrad:
@rawagner why did you change the condition? I think that we have to wait until the shell disappears (and not only until it loses its focus).

rawagner:
When you click on finish in quickfix dialog, often it will cause another dialog to appear and Quickfix dialog is still present and waiting for new dialog to finish.

apodhrad:
test this

apodhrad:
pushed to master

PR #1050, opened 2015-04-24
rawagner_Fix wait condition for waiting till view with specified title is active

mlabuda:
+ add core test plugin
+ rename wait condition


mlabuda:
test this

PR #1053, opened 2015-04-27
ljelinko_Add possibility to execute javascript inside Browser.

rhopp:


rhopp:
retest this

ljelinkova:
* please remove System.out.println from BrowserTest. 
* in log messages, please use present form (evaluate) not continuous form (evaluating)
* why do you catch SWTException in AbstractBrowser.evaluate/execute methods? Display should wrap the exception into CoreException so you should never see SWTException
* why do you return null from evaluate method (this is different in SWT implementation where an exception is thrown). How would you differenciate if the code failed or returned null?
* why do you need wait conditions in the test? If you need to wait for the script to execute, you should put wait into AbstractBrowser

rhopp:
> please remove System.out.println from BrowserTest.

I thought I removed all of them! Damn...

> in log messages, please use present form (evaluate) not continuous form (evaluating)

Ok

> why do you catch SWTException in AbstractBrowser.evaluate/execute methods? Display should wrap the exception into CoreException so you should never see SWTException

The try/catch was placed badly. I should probably catch it in BrowserHandler.

> why do you return null from evaluate method (this is different in SWT implementation where an exception is thrown). How would you differenciate if the code failed or returned null?

This is because I didn't realize, that Display.asyncExec encapsulates  that SWTException, which could lead to null being returned. I also noticed, that I have to refactor execute() method to return boolean.


> why do you need wait conditions in the test? If you need to wait for the script to execute, you should put wait into AbstractBrowser

I can't think of anything I can wait for in AbstractBrowser...

I'll address these issues on monday.

rhopp:
PR updated, BrowserTest passed on all three platforms. Sysouts were removed, exceptions are caught and propagated by syncExec, comments in BrowserHandler were added and execute() is now returning boolean (as it is in org.eclipse.swt.browser.Browser)

ljelinkova:
It seems all good except for the strange WaitUntil in the tests. Do you really need to wait until script is finished explicitly?

rhopp:
test this

rhopp:
test this

rhopp:
BrowserTest passed on all three platforms.

PR #802, opened 2015-04-29
ModifyModulesPage#getConfiguredModules() doesn't work properly

jpeterka:
Following code ends with SWTLayerException
```java
ModifyModulesDialog dialog = server.addAndRemoveModules();
ModifyModulesPage page = dialog.getFirstPage();
if (!page.getConfiguredModules().isEmpty()) {
   page.removeAll();
}
dialog.finish();
```

The problem is that #getConfiguredModules() sets focus to table, so when the #removeAll() is called, the button can't be found because the lookup is not performed on the active shell but on the table.

Stack trace:
```
org.jboss.reddeer.swt.exception.SWTLayerException: No matching widget found
	at org.jboss.reddeer.swt.lookup.WidgetLookup.activeWidget(WidgetLookup.java:206)
	at org.jboss.reddeer.swt.lookup.ButtonLookup.getButton(ButtonLookup.java:39)
	at org.jboss.reddeer.swt.impl.button.AbstractButton.<init>(AbstractButton.java:51)
	at org.jboss.reddeer.swt.impl.button.PushButton.<init>(PushButton.java:62)
	at org.jboss.reddeer.swt.impl.button.PushButton.<init>(PushButton.java:19)
	at org.jboss.reddeer.eclipse.wst.server.ui.wizard.ModifyModulesPage.removeAll(ModifyModulesPage.java:57)
```

### DIFF
--- a/tests/org.jboss.reddeer.eclipse.test/src/org/jboss/reddeer/eclipse/test/wst/server/ui/view/ModifyModulesDialogTest.java
+++ b/tests/org.jboss.reddeer.eclipse.test/src/org/jboss/reddeer/eclipse/test/wst/server/ui/view/ModifyModulesDialogTest.java
@@ -91,4 +91,15 @@ public class ModifyModulesDialogTest extends ServersViewTestCase{
 		modules = server.getModules();
 		assertThat(modules.size(), is(0));
 	}
+	
+	@Test
+	public void testRemoveConfiguredModules() {
+		ModifyModulesDialog dialog = server.addAndRemoveModules();
+		ModifyModulesPage page = new ModifyModulesPage();
+		page.add(PROJECT_1, PROJECT_3);
+		if (!page.getConfiguredModules().isEmpty()) {
+		   page.removeAll();
+		}
+		dialog.finish();
+	}
 }


jpeterka:
Issues seems not to be present anymore. Added this specific tests to cover it for future.

jpeterka:
ready for review

vpakan:
Test testRemoveConfiguredModules() duplicates logic in test addTwo_removeAll() so I suggest just close this issue without pushing changes to repository WDYT?
Both tests are fine now.

jpeterka:
Yes, the differences is mainly usage of getConfiguredModules(). That's why I've added this test to cover this scenario for the future although is's very similar. It's question if we want have test set as small as possible or to cover as much as possible. I was not sure even in time if to close it or add that tests if I just closed it I could have been asked why I didn't rather added the test that covers this precise case. Both objections seems relevant to me but I guess we should have as much coverage as possible in our tests.

vpakan:
I have no problem having many tests but then I would suggest to create test for method getConfiguredModules() instead of duplicating removeAll() testing. Seems to be more reasonable to me?

jpeterka:
This covers exact scenarios that was provided by bug reporter. It was not just the method, I've just said that that makes difference here to other existing tests and hard to say how it can influence rest of the test. So I suggest to push it or close it. It's test modification that seems to be valid as any other.

vpakan:
It does not because API was changed. Also original scenario is not adding modules at least not in snippet in comment.

As you mentioned before only difference to test addTwo_removeAll() is calling method getConfiguredModules() therefore I suggest to update test to test this method otherwise your test makes almost no sense to me IMHO.

jpeterka:
Not sure what you mean by API was changed (except getFirstPage()). That snipped is not complete but its meaning is obvious and some prerequisites like adding modules are necessary as you can see from issue description. This test tests precise scenario nothing more, nothing less. It's not about testing getConfiguredModules() but complete more integrated test case. Original cause was lookup issue which would not be discovered by specific getFirstPage() which you propose anyway.

vpakan:
* yes by API change I meant removing of getFirstPage() method
* what I don't like is just follow falling test scenario instead we should investigate what is problem and then create better test scenario better covering issue. In this case problem was in searching for button "Remove All" which was using table as a root for searching for this button. It was fixed by fixing WidgetLookup.getActiveWidgetParentControl() method which now returns properly shell containing active widget so this is correct issue to test
* new developer reading tests should be able understand their purpose and in this case those two tests mentioned above looks the same. New user will never think about method getConfiguredModules() setting focus to table and causing problem with searching for "Remove All" button afterwards

I'm not willing to push this PR unless test is changed It's against my will but you can ask other gatekeeper to push it I do not have problem with it unless it's not my responsibility.

jpeterka:
I'm not convinced by your argumentation here. I don't think someones potential doubts about tests similarity can beat benefits from test itself that is testing two different API usage within one scenario  although output is the same.  Let's assume you have 2 tests that tests one single user scenario. One is using  Eclipse API and second is doing it via UI. You wish to test both as you provide both API. Test will be the same in general. Situation here is the same and we have an analogy here. One test scenario, two different API that can be used (1. Server.getModules()  and 2. ModifyModulePage.getAvailableModules()).  As this tests is ModifyModulesDialogTest both such a tests operate within the right scope. One think I would actually change now, I would name the test addTwo_removeAll2 instead of unique name. 

I have no problem with removing this test as I'm not myself very rigorous in writing test but it makes me wondering why to refuse additional test. I'm closing this issue as it's not worth to spending more time with it anymore.

PR #210, opened 2014-08-06
SQL Result View implemented

jpeterka:

PR #760, opened 2015-04-29
Internal TreeItem error occurs while reading tree while tree is removed

vpakan:
Found during implementing SQL Result View (#210)
Steps to reproduce: 
STEP: Execute SQL query
STEP: Clean SQL Result View
STEP: Read content of SQL Result View
Reproducible, not 100% but quite easy

I assume TreeItem implementation doesn't count with scenario when some content can be removed during lookup.


vpakan:
test this

vpakan:
It works now correctly therefore closing probably some change introduced by other PR fixed this.

PR #1054, opened 2015-04-30
Make look ups and handlers singleton

mlabuda:
some lookups, handlers does not contain private constructor for creating instance and therefore it's possible to create more instances of such classes although it should not be.


ljelinkova:
Please change also ProgressBarHandler.

PR #1062, opened 2015-05-04
rev_RedDeer Spy requires Java 8

mlabuda:
Currently, the manifest file in org.jboss.reddeer.spy contains

Bundle-RequiredExecutionEnvironment: JavaSE-1.8

It should be JavaSE-1.7


mlabuda:
PR does not change anything else then JRE version of Spy plugin which does not have any test. Because it's compilable and issues of failed jobs on mac and win were not caused by this commit, I suggest to push it without retesting.

PR #1068, opened 2015-05-18
rev_Revise default dependencies for RedDeer test plugin created by wizard

rhopp:
When I create new RedDeer test plugin using wizard (File->New->Other->New RedDeer Test Plug-in), the plug-in has dependencies for org.junit, o.j.r.junit, o.j.r.swt and o.j.r.eclipse.
I would suggest adding at least core and workbench plugins.


rhopp:
retest this

PR #1077, opened 2015-05-21
vpakan_Create snippets for documentation

mlabuda:


mlabuda:
test this

mlabuda:
test this

mlabuda:
Failing tests are not caused by this PR.

vpakan:
https://github.com/jboss-reddeer/reddeer/pull/1077/files#diff-9170ef0c0e5b50ec36375a5d0b7062d8R17 - WORKSPACE_DIRECTORY property should not be initialized to ${WORKSPACE} because it will work only for jobs running on jenkins. Default value should be  '." I guess so it will work also when run manually via maven on local machine.

And also in Javadoc varialble ${WORKSPACE} should be replaced with WORKSPACE_DIRECTORY

PR #889, opened 2015-05-26
ljelinko_Create java requirement

rhopp:


rhopp:
`@JRE` matches any configuration. 
`@JRE(<double>)` matches just exact version of java defined in configuration (for example `@JRE(1.7)` matches configuration with  `<jre:version>1.7</jre:version>`

If there is need to do matching like >=, >, <, ... let me know.

ljelinkova:
- add comment to JREsPreferencePage class
- you may use new FinishButton() instead of new PushButton("Finish")
- getJREs should not return TreeItem but you should create your own type (like in RuntimePreferencePage there is Runtime)
- I would move validateJRELocation directly to the JRERequirement since this is quite specific operation 
- @JRE - add cleanup method

rhopp:
I didn't do bullet "I would move validateJRELocation directly to the JRERequirement since this is quite specific operation" because validateJRELocation is using internal stuff (from package which is not exported). Otherwise it should be ok.

rhopp:
test this

ljelinkova:
All is OK except that JREsPreferencePage still does not have a class comment.

PR #1102, opened 2015-06-04
PropertiesView doesn't wait for tabs to be rendered

jomarko:
If user changes properties of many elements, ProeprtiesView doesn't care about fact, that tabs can be rendered with delay.


vpakan:

apodhrad:
please add a test for the method selectTab()

jomarko:
@apodhrad , do you think that this test [1] is not convenient?

[1]
https://github.com/jomarko/reddeer/blob/master/tests/org.jboss.reddeer.eclipse.test/src/org/jboss/reddeer/eclipse/test/ui/views/properties/TabbedPropertiesTest.java#L71

apodhrad:
@jomarko you are absolutely right, I didn't notice that test.

apodhrad:
test this

apodhrad:
pushed to master

PR #1048, opened 2015-06-04
Static requirement has wrong value injected

ljelinkova:
TestClass1 and TestClass2 extends Template
TestClass1 has req on Server1
TestClass2 has req on Server2

Template has @InjectRequirement static Server s;
Server 2 is injected for both TestClass1 and TestClass2

Workaround: remove static

Lucka will hopefully understand ;)


ljelinkova:
@apodhrad Please review

apodhrad:
pushed to master

PR #1022, opened 2015-06-05
Refactor MarkerIsUpdating condition

ljelinkova:
At the moment, the condition uses "active view" but since the condition can run for a long time, some other view (like Console view) can steal the focus. 

Move the resolving of active view to constructor.


ljelinkova:
test this

ljelinkova:
The build failed on mac after the site was generated so this PR was not the root cause.

vpakan:
Should not be Markers view activated first?

ljelinkova:
No because this condition could be called from different views, e.g. Problems view, Markers view. It is upon caller to activate the right view before the condition is called. 

This is the way it worked before I just moved some part of code from test() method to constructor.

vpakan:
I think we should improve functionality of this condition it should not be dependent on what has happened before it was called. Let's discuss it on meeting

PR #1104, opened 2015-06-11
jpeterka_JBIDE-18772: deploy with "mvn clean verify -Pdeploy-to-jboss.org"

mmalina:
Add profile for deploying RedDeer update site to jboss.org.
More details at https://issues.jboss.org/browse/JBIDE-18772
This is basically a copy of this profile from
https://github.com/jbosstools/jbosstools-build/blob/master/parent/pom.xml

### DIFF
--- a/site/pom.xml
+++ b/site/pom.xml
@@ -70,4 +70,62 @@
 			</plugin>
 		</plugins>
 	</build>
+	
+	<profiles>
+	<profile>
+<id>deploy-to-jboss.org</id>
+<build>
+	<plugins>
+		<plugin>
+			<artifactId>maven-dependency-plugin</artifactId>
+			<executions>
+				<execution>
+					<id>deploy-releng-scripts</id>
+					<goals>
+						<goal>unpack</goal>
+					</goals>
+					<phase>deploy</phase>
+					<configuration>	
+						<artifactItems>
+							<artifactItem>
+								<groupId>org.jboss.tools.releng</groupId>		
+								<artifactId>jbosstools-releng-publish</artifactId>
+								<version>${project.version}</version>
+								<type>tar.gz</type>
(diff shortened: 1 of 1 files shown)


mmalina:
@mickaelistria do you think this will work? Can you review, please?
Currently the update site is here: http://download.jboss.org/jbosstools/mars/snapshots/builds/RedDeer_master/

I'm curious about these variables:
${JOB_NAME}/${BUILD_ID}-B${BUILD_NUMBER} 
I guess they will need to be added to the maven call in the job, right?

mickaelistria:
The patch looks good.
about the variables, Jenkins passes them to Maven without requiring specific effort. See for example the configuraion of a CI job such as jbosstools-livereload to see how "mvn deploy -Pdeploy-to-jboss.org" is invoked.
However, it's not mandatory to use the JOB_NAME and BUILD_ID. For example, you could prefer using ${project.name} instead of ${JOB_NAME} and to skip ${BUILD_ID}. But this would publish to new URLs.

mmalina:
@mickaelistria : Thanks! I think for now we will be happy with exactly the same output directory as we have now.
One more question: The mechanism that makes RedDeer_master/ point to e.g. http://download.jboss.org/jbosstools/mars/snapshots/builds/RedDeer_master/2015-06-05_08-21-04-B623/all/repo/ (via the compositeArtifacts.xml and compositeContent.xml) is separate from the deploy profile? Or how does that work? Because the rsync.sh only publishes to the subdirectory, no?

mickaelistria:
the rsync.sh also takes care of updating the composite in RedDeer_master/ .

mmalina:
The linux build passed and since this does not affect any testing (it's just a snipped in the site/pom.xml to publish the update site) that should be fine I guess.

PR #1109, opened 2015-06-17
Change publish url mars -> snapshot

mmalina:
It turns out RedDeer is not on mars yet, so
it makes more sense to publish to an url
with luna in it.

### DIFF
--- a/site/pom.xml
+++ b/site/pom.xml
@@ -90,7 +90,7 @@
 							<artifactItem>
 								<groupId>org.jboss.tools.releng</groupId>		
 								<artifactId>jbosstools-releng-publish</artifactId>
-								<version>${project.version}</version>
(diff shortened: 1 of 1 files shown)


mmalina:
Once this is done, we can finally attempt to publish the site using the maven profile instead of using the publish.sh script directly in the job.

jpeterka:
Please make master update site name eclipse version agnostic

mmalina:
Well, at this point I am only trying to move the jenkins job to use the new publish mechanism. Until that is done, I don't want to change anything. Also, the url is included in both the job and the pom. So right now they are not in sync.

mmalina:
ok, so I updated the config to remove mars or luna, let's see if that works. but I also needed to hardcode the plugin version there. please push this ASAP, so I can finally get it to run.

PR #1110, opened 2015-06-18
JBIDE-18772 use consistent path for publishing snapshot builds

nickboldt:


vpakan:

mmalina:
I don't think this is gonna work, @nickboldt . First Max needs to to talk to Jirka and see if they find an agreement on this topic. And second, if anything, it should be luna at this point (see the root pom, it links to eclipse luna repo). Also, see #1109

nickboldt:
So, since we don't have a /luna/snapshots/ folder (this path concept was added for Mars) why not publish to the old location? 

http://download.jboss.org/jbosstools/builds/staging/RedDeer_master/
or
http://download.jboss.org/jbosstools/builds/staging/RedDeer_luna/ 

Incidentally, if Red Deer master is still on Luna, why do you have two folders there?

mmalina:
We were fine with the old url. Only recently we noticed that the update site on that url has not been updated since April 30. And since the new url was already set up in jenkins, I just updated all places to consume that. Sure, we could move to the old url. I'm not exactly sure, but I think RedDeer_luna is the result of the the luna job, which uses the previous version of RedDeer - 0.7.x. My assumption was that master = mars, but it turns out it's not the case.

nickboldt:
PR updated to use old path: builds/staging/RedDeer_master/all/repo/

Note that RedDeer_master currently contains 0.8.0 and RedDeer_luna contains 0.6.2:

http://download.jboss.org/jbosstools/builds/staging/RedDeer_master/all/repo/features/ 0.8.0.201504301324
http://download.jboss.org/jbosstools/builds/staging/RedDeer_luna/all/repo/features/ 0.6.2.201410211408

mmalina:
ok, I'm fine with this, but whoever pushes this, keep in mind that the jenkins job needs to be modified also - feel free to ping me for that.

nickboldt:
What in the job needs to be changed? The destination publishing path is only configured in the site's pom.xml.

mmalina:
@nickboldt this:
if [[ ${skipRevisionCheckWhenPublishing} == "true" ]] || [[ $(. ${WORKSPACE}/sources/util/checkLatestPublishedSHA.sh -s ${WORKSPACE}/sources/site/target/repository/ -t http://download.jboss.org/jbosstools/snapshots/builds/${JOB_NAME}/latest/all/repo/) == "true" ]]; then ...

It's not critical. I guess if you don't change that it will just always publish which is not a biggie.

nickboldt:
Yes, good point. Tha twould need to be changed each time the pom's publishing location was updated.

You could consider externalizing that path to a maven variable, then you could use a Jenkins job param to define the path, pass that in to the schell script, AND pass it to maven, too. Then it would only need to be changed in ONE place.

vpakan:
OK to test

vpakan:
After discussion we agreed that this URL http://download.jboss.org/jbosstools/snapshots/builds/RedDeer_master/ is fine for us to publish latest builds of RedDeer master to. Please do not add Eclipse such Luna or Mars version to this URL we would like to have one stable URL pointing always to RedDeer master update site bits.

Are you ok with thisf @nickboldt ? If so I will close this issue without pushing it.

mmalina:
@nickboldt alternatively you can suggest a different url, but one that is eclipse version agnostic. What this PR is suggesting atm is moving back to the old path (with /staging/), but you say this is for luna only and we're gonna move to mars soon (for the build - we do support both luna and mars for the consuming projects).

jpeterka:
Stable sites currently look like this http://download.jboss.org/jbosstools/updates/stable/luna/core/reddeer/0.6.0/ , not much similarity there. We should suggest overall RedDeer location schema approach. Probably we should update stable sites schema too. Created #1111

nickboldt:
Sorry, you can't have an Eclipse-versionless URL. Because what will you do when you suddenly change Red Deer so it won't work on Luna? Where would you put it? Into a /mars/ folder? But then would you have to MOVE the stuff that wasn't already in /luna/ to be consistent? Or would "no version" mean Luna and "/mars/" mean "mars-only" ... or "mars and neon" ?

See, it gets confusing. :D 

We had this problem like 5 years ago with JBoss Tools -- and we solved it by always having a different URL every year. If your bits for Luna still work on Mars, that's great. Just tell people that.

Similarly, a project like Atlassian's JIRA connector for Eclipse uses a URL like this:

http://update.atlassian.com/atlassian-eclipse-plugin/rest/nightly/e3.7

but it works fine with Eclipse 3.8, 4.3, 4.4, ...

Or SpringIDE:

http://dist.springsource.com/release/TOOLS/update/e4.4/ (works with Luna AND Mars).

If you really want a "vanity URL" you can achieve that like this:

http://download.jboss.org/jbosstools/updates/development/ (note the README and composite*.xml which point to the /mars/ equivalent folder)

But you still PUBLISH to the correct place... you just make it easier for people to find it under a URL with one less path segment.

nickboldt:
For consistency, I've moved updates/development/core/reddeer/0.8.0.M1 into http://download.jboss.org/jbosstools/updates/development/luna/core/reddeer/ and removed duplicate 0.8.0.m1 (why lowercase?) folder.

Maybe you want a symlink from /updates/development/reddeer to the latest release in /updates/development/luna/core/reddeer/ ? (and another in /updates/stable/reddeer/ linking to /updates/stable/luna/core/reddeer/ ?

nickboldt:
So, I've updated your job to use a publishingPath variable, which you can set at build time to push bits to the Luna path or the Mars one. 

https://jenkins.mw.lab.eng.bos.redhat.com/hudson/job/RedDeer_master/jobConfigHistory/showDiffFiles?timestamp1=2015-06-23_23-56-14&timestamp2=2015-06-24_12-14-40

Also updated this PR to show the two options (and default to the Luna one for now).

mmalina:
Thanks a lot, @nickboldt . I thought there was one thing missing - the checkLatestPublishedSHA.sh call should also use this new variable. But actually not (correct me if I'm wrong) - this one points to latest/ at the end, whereas the new ${publishingPath} variable contains the build number. So that means there are still two places to update when needed, but at least it's all in the job config now :)

So +1 to merge this.

nickboldt:
Job updated to check publishingPathLatest, which is publishingPath but s/buildID+number/latest/.

Can we merge this now?

PR #1115, opened 2015-06-25
JBIDE-18772 use consistent path for publishin...
...g snapshot builds

nickboldt:
JBIDE-18772 use builds/staging/RedDeer_master/all/repo/ for current Red Deer 0.8.0 nightly builds against Luna; later move to /mars/ path

externalize publishingPath so it can be passed in from Jenkins

since Red Deer now builds on Mars, we can remove the old Luna convention

### DIFF
--- a/site/pom.xml
+++ b/site/pom.xml
@@ -12,6 +12,10 @@
 		<version>0.8.0-SNAPSHOT</version>
 	</parent>
 
+	<properties>
+		<publishingPath>mars/snapshots/builds/${JOB_NAME}/${BUILD_ID}-B${BUILD_NUMBER}/all/repo/</publishingPath>
+	</properties>
+
 	<build>
 		<plugins>
 			<plugin>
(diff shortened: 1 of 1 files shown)


mmalina:
+1, @jpeterka please merge.

site/pom.xml
```diff
@@ -12,6 +12,10 @@
 		<version>0.8.0-SNAPSHOT</version>
 	</parent>
 
+	<properties>
+		<publishingPath>mars/snapshots/builds/${JOB_NAME}/${BUILD_ID}-B${BUILD_NUMBER}/all/repo/</publishingPath>
```
maxandersen: for this to have effect the RedDeer job needs changing to reddeer to follow the conventions used everywhere else.

nickboldt: Not technically true (case conventions do not define functionality, only stylistic preference), but I agree that renaming the job to "reddeer_master" or "jbosstools-reddeer_master" would be more consistent w/ the rest of the JBT projects.

maxandersen: It is technically true since URL's are case sensitive on some systems thus it does have technical meaning and functional differences.

nickboldt: I'll change the job name if you merge the commit. PLEASE? /me has been waiting for days now. :(

jpeterka:
pushed to master

site/pom.xml
maxandersen: none of this is under jbosstools so i can't push. and the jobname could be changed without this fix afaics.

but jiri merged so also should be fine to go forward.

mmalina:
I created a PR for integration tests to update to this new master URL:
https://github.com/jbosstools/jbosstools-integration-tests/pull/1072

PR #1116, opened 2015-06-26
Mars_ViewHandler.setFocus issue

ljelinkova:

PR #1131, opened 2015-07-08
jpeterka_Mars_ViewHandler.setFocus issue

ljelinkova:
This is a duplicate issue since Github <-> Jenkins integration was broken on tha original #1116

### DIFF
--- a/plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/CTabItemHandler.java
+++ b/plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/CTabItemHandler.java
@@ -161,6 +161,19 @@ public class CTabItemHandler {
 			}
 		});
 	}
+	
+	/**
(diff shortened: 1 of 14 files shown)


apodhrad:
- What's the reson of adding new method 'activeWidgets' which is very similar to the original one?
- Test ViewWithTitleIsActiveTest.testConsoleView is still failing
- In AbstractView.java don't break current formatting (find '.performAction')

jpeterka:
Just @Ignore testFocusOnInstantiatedView instead of removal, can be used later on when upstream bug is fixed.

ljelinkova:
@jpeterka  OK
@apodhrad 
- the main reason is that it can return all widgets (not just one) and there is no wait time (when calling original one there is wait time of 10 seconds) - thus it is more suitable for simple check if something is present or not (like if view ctab is open)
- ViewWithTitleIsActiveTest should be ignored
- OK

PR #1126, opened 2015-07-08
Improve algorithm for find a labeled widget

apodhrad:
Current implementation of WidgetHandler.getLabel() doesn't work if a label is added after the widget, possible solution could be to add the following

```java
	@Override
	public boolean matches(Object obj) {
		if (!(obj instanceof Text || obj instanceof Combo)) {
			return false;
		}

		Control control = (Control) obj;
		Control parent = control.getParent();

		java.util.List<Widget> children = WidgetResolver.getInstance().getChildren(parent);
		for (Widget w : children) {
			if (w instanceof Label || w instanceof CLabel) {
				CLabel cLabel = (CLabel) w;
				Object layoutData = cLabel.getLayoutData();
				if (layoutData instanceof FormData) {
					FormData formData = (FormData) layoutData;
					if (control.equals(formData.left.control) || control.equals(formData.right.control)) {
						if (matcher.matches(cLabel.getText())) {
							return true;
						}
					}
				}
			}
		}

		return false;
	}
```


jpeterka:
retest this please

apodhrad:
fyi it was already tested ;-)

PR #1114, opened 2015-07-08
Allow to create screenshots even in code

mlabuda:
to use like ScreenshotCapturer.capture(fileName) or similar.


mlabuda:
test this

mlabuda:
test this

jpeterka:
Please rebase and resolve conflicts and provide documentation or create doc issue.

mlabuda:
#1143  doc issue

jpeterka:
Some more findings:
1) when captureScreenshot is called without extension  java.lang.StringIndexOutOfBoundsException is thrown and it's not handled properly. Argument should be checked and error should be handled better
2) javadoc for captureScreenshot refers "File name has to contain file extension (e.g. .png)" but actually when other extension is used (like .tiff) it's png file anyway. 
3)default file path should for custom screenshots should be target/screenshots or target at least.

PR #902, opened 2015-07-09
GEFEditor - adding element to subprocess of BPMN2 diagram

apodhrad:
There can be nested elements in BPMN2 diagram. Typical example:
StartEvent -> SubProcess -> EndEvent
SubProcess:StartEvent -> SubProcess:ScriptTask -> SubProcess:EndEvent

Method addToolFromPalette() of GEFEditor detects added element, by hanging listener on DiagramShapeEditPart. It is needed, to listener be hanged on direct parent of added element.


apodhrad:
The listener which detects new edit part was added to all existing edit parts.

PR #269, opened 2015-07-09
Create a condition for detecting if JUnit run has finished

apodhrad:


apodhrad:
The new test JUnitRunTest breaks another test JUnitViewTest if test order is as follows:
1. JUnitRunTest - test passes but the results are kept
2. JUnitViewTest - this test expects a JUnit view without any results

apodhrad:
There is the following problem
- a status doesn't starts with "Finished" if JUnit view is not opened and you start the test for the first time

As a workaround I just open the JUnit view before the test.

apodhrad:
I solved the problem as follows
1. Check whether status is e.g. 2/2
2. Then, wait until there is no running job related to JUnit

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/condition/JobIsRunning.java
```diff
@@ -74,8 +96,13 @@ public boolean test() {
...
 			}
 
-			if (job.isSystem() || job.getState() == Job.SLEEPING) {
-				log.debug("  job '%s' is system job or not running, skipped", job.getName());
+			if (skipSystemJobs && job.isSystem()) { 
+				log.debug("  job '%s' is a system job, skipped", job.getName());
+				continue;
+			}
+			
+			if (job.getState() == Job.SLEEPING) {
```
rawagner: Shouldn't there be else ? For system job which is sleeping I will get two log messages which is not necessary

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/core/resources/ProjectItem.java
```diff
@@ -55,6 +59,46 @@ public ProjectItem(TreeItem treeItem) {
...
+	public void runAsJUnitTest() {
+		runAsJUnitTest(TimePeriod.LONG);
+	}
+	
+	/**
+	 * Runs project as JUnit Test waits until it is finished in a given timeout.
+	 * 
+	 * @param timeout Timeout
+	 */
+	public void runAsJUnitTest(TimePeriod timeout) {
```
rawagner: You should mention in javadoc that console activation is being turned off during this action

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/condition/JobIsRunning.java
apodhrad: That will never happen since there is CONTINUE statement.

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/core/resources/ProjectItem.java
apodhrad: I don't think it is necessary to mention it because the logic is completely encapsulated into the method (the settings for console activation are set back at the end of the method).

apodhrad:
Done.

PR #1136, opened 2015-07-10
apodhrad_Mars_TabbedPropertiesTest

ljelinkova:


ljelinkova:
Retest this

ljelinkova:
Mac job failed again but it seems to be a norm now so I am marking this issue for review anyway

apodhrad:
The following issues were fixed
1. Opening an editor may take a while - added a wait condition
2. Shell title is changed when the editor is opened - fixed in setFocus() method

Lets think about what can be wrong with the change in method setFocus().

apodhrad:
I don't see any issue with the change number 2; pushed to master

PR #362, opened 2015-07-14
Diagnostic information provided after test failure

apodhrad:
[INFO] Opening "Preferences" shell
[ERROR] Shell not found 
[DUMP] Available shells: "Usage Reporting"
This is general tracking issue, it will be improved step by step


apodhrad:
The diagnostic information are provided only in the exception message, do we also need it in the log?

Issue #1039, opened 2015-04-15
updatesite - Recorder is still in UI section


rawagner:

vpakan:
Fixed by #1146

PR #1146, opened 2015-07-14
rev_Add Log Parser Tool

vpakan:


vpakan:
Fixes #1039

vpakan:
Verification job results are irrelevant for this issue

vpakan:
Documentation isuue has been created #1147

jpeterka:
1) editor content is not refreshed when file content is changed
2) consider adding  isEditorInputReadOnly() and isEditorInputModifiable() for read-only editor
3) why are view tool items duplicated (once as toolbar and once as drop down toolbar menu items)?
4) javadoc for multiple public members is missing

vpakan:
1) that is feature because for long files it takes long time to parse and load log so it's always up to user ot decide when to do it
2) can do it but don't think it's correct because EditorInput is editable but editor is not
3) just because default example for new view has it the same way it's no problem to remove it but  for me it's ok if it's there
4) this is not API so javadoc is not needed as other tools doesn't have it either

jpeterka:
1) I agree with that but when I click in the Log Parser View I would expect up-to-date version. I don't think it should watch for changes but user force refresh should work
2) ok
3) it's minor issue but still it's duplicated, If anyone duplicate all the toolbar items Eclipse would be pretty weedy
4) Proper public javadoc is good practice in general. If we miss it somewhere else we should add it I think. Yes, I can also find many bad examples of missing javadoc even in Eclipse platform but we should rather look for good examples.

vpakan:
Points 1. and 3. fixed. Javadoc will be discussed later

vpakan:
test this

vpakan:
Compilation was successful test results are irellevant

PR #1144, opened 2015-07-15
Snippet wiki generator generates new lines every time jenkins job is run

mlabuda:
For example see Database-Requirement-Extension.md. After each snippet in that file, there are ~40 empty lines.


mlabuda:
This PR does not depend on verification job.

mlabuda:
Code is compilable (on linux). Job has been terminated, no need to let tests run in this case.

PR #1156, opened 2015-07-21
rev_Use less restriction in casting in CellEditor

apodhrad:
It is enough to cast to ViewerColumn instead of TableViewerColumn.


apodhrad:
this is very small and save change, so we can include it into 0.8.0

PR #1157, opened 2015-07-23
rev_Method addToolFromPalette doesn't work if there is a wizard before adding a component

apodhrad:
Use case:
1. Add a tool from palette
2. A wizard pops up to set some options
3. After finishing the wizard the component is added

The problem here is that the method addToolFromPalette immediately waits for the new component.


apodhrad:
Very small change here; just extracting a part of the method to another method which is protected and can be overridden if there is a wizard.

PR #1179, opened 2015-08-06
DeleteUtils.forceProjectDeletion() catches wrong exception

apodhrad:
```java
public static void forceProjectDeletion(Project project, boolean deleteFromFileSystem) {
	try {
		project.delete(deleteFromFileSystem);
	} catch (EclipseLayerException ele) {
		log.debug("Delete project '" + project.getName() + "' via Eclipse API ");
		org.jboss.reddeer.direct.project.Project.delete(project.getName(), deleteFromFileSystem, true);
	}
}
```

the code above catches EclipseLayerException but the try-block throws SWTLayerException with higher probability.

### DIFF
--- a/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/utils/DeleteUtils.java
+++ b/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/utils/DeleteUtils.java
@@ -5,9 +5,9 @@ import org.eclipse.swt.widgets.Shell;
 import org.hamcrest.BaseMatcher;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
+import org.jboss.reddeer.common.exception.RedDeerException;
 import org.jboss.reddeer.common.logging.Logger;
 import org.jboss.reddeer.eclipse.core.resources.Project;
-import org.jboss.reddeer.eclipse.exception.EclipseLayerException;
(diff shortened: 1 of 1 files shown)


mlabuda:
It's not true. handleDeleteResourcesShell and DeleteUtils.handleDeletion are only methods where could SWTLayerException occurs but it is catched and returns boolean. So finally only EclipseLayerException can be thrown in case of not working deletion either via context menu or shell menu.

apodhrad:
And what's wrong with catching RedDeerException? This will cover all exceptions we need, wdyt?

vpakan:
@mlabuda @apodhrad I vote for RedDeer exception because Project is coming from core RD plugin and therefore we cannot guarantee what Exceptions can project.delete() method throw in future.

And watch long lasting PRs please here was no response for more than 2 months.

vpakan:
test this

mlabuda:
Project is in eclipse plugin, but name of package is unfortunate org.jboss.reddeer.eclipse.core.resources. Currently delete can throw in most of the cases EclipseLayerException. But it's true there could be also different exception - e.g. SWTLayerException in case of non-existence of a context menu Delete (if something in Eclipse would get broken and this context menu item would disappear). Of there could occur unpredictable issues. We are not encapsulating such issues into EclipseLayerException. So it would fail on different exception and therefore project would not be deleted but forced delete from delete utils would not be performed because of it. Taking all above into consideration, I think, it would be better to catch RedDeerException.

Issue #124, opened 2013-04-17
UserGuide - RedDeer installation


jpeterka:
Add a chapter about RedDeer installation into UserGuide (https://github.com/jboss-reddeer/reddeer/wiki/User-Guide)

Please follow https://github.com/jboss-reddeer/reddeer/wiki/UserGuideTemplate for more information about User Guide

PR #1108, opened 2015-08-11
Add logic for known issues

mlabuda:
I would like to have something like this

```java
@Test(knownIssues = {REDDEER-1234})
```

and when this test fails then it will throw an exception saying

"This is known issue, please see REDDEER-1234"
"caused by ..."

### DIFF
--- a/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/properties/RedDeerProperties.java
+++ b/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/properties/RedDeerProperties.java
@@ -30,6 +30,8 @@ public enum RedDeerProperties {
 	 */
 	CONFIG_FILE("reddeer.config", (String) null),
 	
+	ISSUE_TRACKERS("reddeer.issue.trackers", (String) null),
+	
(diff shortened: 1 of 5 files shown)


vpakan:
Lets discuss these comments on meeting:
* annotation could not be set for line so it's not possible to check if test failed on proper expected line within test method therefore I don't want to have these feature
* new exception containing known issues should be KnownIssueException not just Throwable
* what to do when method has known issue but did not fail? Probably it should fail saying something like: "This test should fail because it's marked as known issue... "

jpeterka:
I completely agree with Vlado. If you want to handle known issue I suggest to use mechanism we are using in JBT:

```java
try {
  // code producing issue
} catch(SpecificExceptionType e) {
  fail("Something happened, known issue(s) #1234")
}
```

If you want to to have some mechanism we can implement some API to translate REDDEER-1234 to 
https://github.com/jboss-reddeer/reddeer/issues/124 . Usage could look like this

```java
fail(IssueResolver.getMessage("REDDEER-1234");
```
that would produce for example:
Test failed possibly for known issue(s) are REDDEER-1234, see https://github.com/jboss-reddeer/reddeer/issues/1234

jpeterka:
Rejected

PR #1210, opened 2015-08-25
JBIDE-20194 fix license, copyright, and provi...
...der-name

nickboldt:


vpakan:
Rejected.
RedDeer is not part of JBT and therefore has no dependencies to JBT so this PR is not compilable. RedDeer is going to be Eclipse project.

nickboldt:
Fair enough but you could still update the license to be the correct one. I guess if you'll be at Eclipse.org within a year then it's less urgent, however.

PR #1133, opened 2015-08-26
rev_Mark ButtonWithTextIsEnabled as deprecated (WidgetIsEnabled)

mlabuda:
Because it takes as parameter Button not text.


mlabuda:
test this

PR #971, opened 2015-08-26
Wait for Shell with empty title doesn't work properly

jpeterka:
see this construction
```java
new WaitUntil(new ShellWithTextIsActive(""));
```
it timeouts, probably some bug in matcher or lookup


jpeterka:
Cannot replicate this issue anymore, added some tests to keep it ok.

PR #1211, opened 2015-08-26
Wrong line is bold in SpyView

mlabuda:
http://snag.gy/9j02J.jpg


mlabuda:
Failed job on windows is not affected by this commit. I think this can be reviewed and merged.

PR #1218, opened 2015-08-27
rawagner_Implement click() method within TreeItemHandler

vpakan:


plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/TreeItemHandler.java
```diff
@@ -381,6 +382,60 @@ public void run() {
...
+	 * Creates Mouse Event for TreeItem
+	 * @param swtTreeItem
+	 * @param x
+	 * @param y
+	 * @param button
+	 * @param stateMask
+	 * @param count
+	 * @return
+	 */
+	private Event createMouseEvent(TreeItem swtTreeItem, int x, int y, int button, int stateMask, int count) {
```
rawagner: Might be safer to use TreeHandler#createEventForTree at first and then add other needed info. 
createEventForTree specifies other event attributes such as type and item which might be missing depending on listener implementation

rawagner:
Can you also add test for this ?

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/TreeItemHandler.java
vpakan: I don't understand. This method is creating mouse event for tree item not event for tree. What would be safer?

vpakan:
Any ideas about test scenario? I don't know how to validate if mouse click on tree item actually happened. Defining listener and evaluating event data within listener doesn't make sense.

It would be tested in JBT integration tests using more complex test scenario

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/TreeItemHandler.java
rawagner: I didn't notice that event.widget has to be treeItem. Then add event.item and event.type to this method

rawagner:
Test - ShellMenu File->New->Other... click on any treeItem and check if it was selected would be enough to have at least something.

vpakan:
Ready for push

PR #102, opened 2015-08-28
Possibility to define when requirement extension should be fulfilled

mlabuda:
Ability to define when requirement extension is executed

```java

Transparent requirement usage:
class
test method

@CleanWorkspace
@BeforeClass	// default behavior when used before class statement
public static prepare() {}

@AfterClass - performs fullfill after class (cleanup)
@Before - performs before each
@After - perform after each test (cleanup)

@Test
@CleanWorkspace // dependency performed only for particular test
public vod 

@CleanWorkspace /// this will be performed once at the beginning, same as when used @BeforeClass
 class { ...

```


jpeterka:
RD mtg decision: this will not be supported

PR #1231, opened 2015-08-31
Handle preference page changed/dialog ok

rawagner:


rawagner:
retest this

jpeterka:
RD mtg decision: this is not needed

PR #1230, opened 2015-09-01
jpeterka_org.jboss.reddeer.swt.test.EditorBarTest.workbenchToolBarTest is failing

rhopp:
http://machydra.brq.redhat.com:8080/job/RedDeer_master_matrix/jdk=sunjdk1.7,label=firebird/668/testReport/junit/org.jboss.reddeer.swt.test/EditorBarTest/workbenchToolBarTest_default/?

Error Message

Timeout after: 10 s.: shell with text matching"New" is active
Stacktrace

org.jboss.reddeer.common.exception.WaitTimeoutExpiredException: Timeout after: 10 s.: shell with text matching"New" is active
	at org.jboss.reddeer.common.wait.AbstractWait.timeoutExceeded(AbstractWait.java:159)
	at org.jboss.reddeer.common.wait.AbstractWait.wait(AbstractWait.java:112)
	at org.jboss.reddeer.common.wait.AbstractWait.<init>(AbstractWait.java:77)
	at org.jboss.reddeer.common.wait.AbstractWait.<init>(AbstractWait.java:53)
	at org.jboss.reddeer.common.wait.AbstractWait.<init>(AbstractWait.java:41)
	at org.jboss.reddeer.common.wait.WaitUntil.<init>(WaitUntil.java:25)
	at org.jboss.reddeer.swt.test.EditorBarTest.prepare(EditorBarTest.java:37)

### DIFF
--- a/tests/org.jboss.reddeer.swt.test/src/org/jboss/reddeer/swt/test/EditorBarTest.java
+++ b/tests/org.jboss.reddeer.swt.test/src/org/jboss/reddeer/swt/test/EditorBarTest.java
@@ -34,7 +34,7 @@ public class EditorBarTest {
 		// new DefaultShell("Spring Tool Tips").close();
 		new WorkbenchShell();
 		new ShellMenu("File","New","Other...").select();
-		new WaitUntil(new ShellWithTextIsActive("New"));
+		new DefaultShell("New");
 		new DefaultTreeItem("General","Project").select();
 		new PushButton("Next >").click();
 		new WaitUntil(new ShellWithTextIsActive("New Project"));


jpeterka:
Wouldn't be safer to change it to ?
```java
new WaitUntil(new ShellWithTextIsAvailable("New"));
new DefaultShell("New");
```

rawagner:
@jpeterka  its basically the same. DefaultShell will wait for 5sec for "New" shell to appear.

rhopp:
Nope... That wait was the thing which was causing the trouble. After the "New" shell was opened, on Firebird no shell had focus, thus the wait failed. `new DefaultShell("New");` waits for the shell to appear even though it appears unfocused and activates it afterwards.

rhopp:
So I guess we can say, that using `WaitUntil(ShellWithTextIsActive)` for waiting for shell to open could be considered bad practice as it can fail on some platforms and instead of that we should use `new DefaultShell("text")`.
Note, that using `ShellWithTextIsActive` condition with WaitWhile for waiting for shell to close is OK.

jpeterka:
Ok, let's keep it like this. There is timeout 10s by default for this which is more than enough for this case. ShellWithTextAVAILABLE condition would be too much, you're right.

jpeterka:
Please add fixes #no next time into commit message

PR #1246, opened 2015-09-14
rev_Add support to analyze that screenshot is blank

jpeterka:
Sometimes screenshot taken during test is blank (white or black usually). This can be caused by several factors (wrong X display, broken GUI session, etc.).  RedDeer should provide API to detect this

```java
ScreenshotCapturer sc = ScreenshotCapturer.getInstance();
String filePath = sc.captureScreenshot("test.png");
ImageAnalyzer ia = new ImageAnalyzer.getInstance();
ia.isImageBlank(filePath);
```


apodhrad:
pushed to master

PR #1248, opened 2015-09-21
Force focus finer issues handling 

jpeterka:
Sometimes forceFocus doesn't work and it's not obvious what happens. So finer state handling should be added.


jpeterka:
retest this please

PR #1250, opened 2015-09-21
clean up update site html and add missing ver...
...sion for org.jboss.tools.tycho-plugins:repository-utils; also move to use tycho 0.23.1

nickboldt:


vpakan:
Verification job failed. Please check.

Be aware this is not JBT build so it's not using JBT parent pom maybe it's the issue?.

nickboldt:
Worked fine when I built it locally using Maven 3.2.5 and JDK 8

Failure is because you're using JDK7 instead of 8.

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:0.23.1:test (default-test) on project org.jboss.reddeer.common.test: Execution default-test of goal org.eclipse.tycho:tycho-surefire-plugin:0.23.1:test failed: Invalid extraRequirement jar:jmock-legacy:2.5.1: Unknown artifact type "jar" -> [Help 1]
```

If you need to build w/ JDK 7, you can revert my change on line 12 and stick with an older version of Tycho 0.22 (instead of moving up to 0.23.1).

vpakan:
Yes we want to build with Java 1.7.

nickboldt:
You could set BREE = 1.7 in your MANIFEST.MF files and still build w/ 1.8. Or use my amended PR.

vpakan:
I'm waiting for verification job to finish to confirm your amended commit works fine with Java 1.7

PR #1256, opened 2015-09-28
Add Build Paths property page

vpakan:


vpakan:
test this

ljelinkova:
- please update javadoc for BuildPathsPropertyPage constructor
- correct typo in removeVariable (globaly -> globally)
- the test does not contain any assert - you could add method getLibraries to the BuildPathsPropertyPage and use it in the test
- some of the functionality (the global preferences) should be extracted to separate class because it is a PreferencePage. Thus it can be used as preference page and from project's property page.

PR #1201, opened 2015-10-02
Make the timeout configuratble via RD properties

ljelinkova:
Timeout multiplication property is meant here.

### DIFF
--- a/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/TimePeriod.java
+++ b/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/TimePeriod.java
@@ -1,6 +1,6 @@
 package org.jboss.reddeer.common.wait;
 
-import org.jboss.reddeer.common.wait.TimePeriod;
+import org.jboss.reddeer.common.properties.RedDeerProperties;
 
 /**
  * Represents the time period for how long the user operation might last. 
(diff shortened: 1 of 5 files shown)


ljelinkova:
Doc issue created https://github.com/jboss-reddeer/reddeer/issues/1259

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/TimePeriod.java
```diff
@@ -42,7 +44,20 @@ private TimePeriod(long seconds) {
...
+
+		if (FACTOR == 1) {
+			return seconds;
+		}
+
+		// do not exceed max value of Long
+		if (seconds >= Long.MAX_VALUE / FACTOR) {
+			return Long.MAX_VALUE;
+		} else {
+			return Math.round(seconds * FACTOR);
```
vpakan: Use this expression Math.round((double)seconds * factor) otherwise round method will return Integer.MAX_VALUE for each input greater then Integer.MAX_VALUE.
It's because input parameter for round method is evaluated to float and therefore it returns integer instead of long.

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/TimePeriod.java
```diff
@@ -62,4 +77,8 @@ public static TimePeriod getCustom(long seconds) {
 	public String toString() {
 		return "Time period " + seconds + " s.";
```
vpakan: Add factor variable to toString() method

PR #1261, opened 2015-10-06
Add timeout - Correct timer waiting for widget to be available in RepoConnectionDialog

ldimaggi:


vpakan:
This commit is already in repo. PR rejected

Issue #794, opened 2014-08-28
Complete tests of swt widgets


ljelinkova:
Every swt widget should have at least one test in swt.test plugin. The following is (incomplete) list of widgets without tests:

DefaultList

mlabuda:
Another missing tests:
* Label
* Menu tests (ToolbarMenu and ShellMenu)
* Shell test (DefaultShell and WorkbenchShell)
* Toolbar tests (DefaultToolBar, ViewToolBar, ViewToolItem)

jrichter1:
So, what exactly is there to test? I looked through the existing tests and all of the tests listed as missing are present in the o.j.r.swt.test package, just not each in a separate class.

These test classes are present:
LabelTest class for labels
MenuTest class for the menus
ShellTest class for the shells
ToolBarTest class for the toolbars

Is there something I'm missing, or this issue is no longer relevant?

ljelinkova:
If tests are already present, please move them to the appropriate packages and rename them to correspond to the tested class name (e.g rename ShellTest to WorkbenchShellTest). 

The aim is to be able to find the tests for every widget easily - by the package and the name of the widget.

PR #1262, opened 2015-10-07
Move widget tests to their appropriate package and class

jrichter1:
Splits some classes that grouped different widget implementation tests into separate classes and moves the into the appropriate packages. Fills in the missing test classes from #794.


jrichter1:
test this

PR #1257, opened 2015-10-21
jpeterka_Screenshot not created when screenshot directory does not exist

mlabuda:
11:55:17.333 INFO [WorkbenchTestable][Requirements] Fulfilling requirement of class org.jboss.reddeer.requirements.openperspective.OpenPerspectiveRequirement
11:55:17.335 ERROR [WorkbenchTestable][Requirements] Capturing screenshot failed because of following error: Screenshot could not be created, because required folders where to store screenshot has not been created
11:55:17.335 ERROR [WorkbenchTestable][Requirements] Error was caused by: 
null
11:55:17.335 INFO [WorkbenchTestable][ShellHandler] Closing all shells...
11:55:17.341 DEBUG [WorkbenchTestable][AbstractWait] Waiting until active shell exists...
11:55:17.342 DEBUG [WorkbenchTestable][AbstractWait] Waiting until active shell exists finished successfully

Check log of job RedDeer_verification_matrix/2236/jdk=sunjdk1.7,label=stable-linux/consoleText

### DIFF
--- a/plugins/org.jboss.reddeer.junit/src/org/jboss/reddeer/junit/screenshot/ScreenshotCapturer.java
+++ b/plugins/org.jboss.reddeer.junit/src/org/jboss/reddeer/junit/screenshot/ScreenshotCapturer.java
@@ -73,8 +73,8 @@ public class ScreenshotCapturer {
 			
 			captureScreenshot(path + name + ".png");
 		} else {
-			throw new CaptureScreenshotException("Screenshot could not be created, "
-					+ "because required folders where to store screenshot has not been created");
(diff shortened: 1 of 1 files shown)


mlabuda:
Changed behaviour of screenshot capturing. Issue was incorrect logging message. It was not in non-existing directories but RedDeer property whether a screenshot should be captured on failure or not was set to false (or better, it was not true and thus screenshot was not created.). In this case I think its better to log warning info that screenshots are not capturing instead of throwing an exception.

mlabuda:
test this

PR #1279, opened 2015-10-26
Remove "New" from RedDeer wizards title

jpeterka:


ljelinkova:
Issue was fixed by #1066

Issue #1220, opened 2015-08-28
Migrate RedDeer doc part 1


mlabuda:
Migrate following topics:
- Write Minimal RedDeer Test
- Running RedDeer Tests
- Write your own RedDeer API
- Execution Parameters
- Run RedDeer Tests in Jenkins Environment

mlabuda:
Closing. Decided to not migrate at the moment. Keep in .md format.

PR #1283, opened 2015-10-26
mlabuda_Migrate Write Minimal RedDeer Test

jpeterka:
sub of #1220


mlabuda:
rejecting, we decided not to migrate at the moment

PR #1155, opened 2015-11-04
Implement CCombo

apodhrad:


vpakan:
Please move related logic from ComboHandler.java to separate CComboHandler.java

rawagner:
+1 We are using C*Handlers for C*Widgets already.

PR #1215, opened 2015-11-04
rev_AbstractWidget should have isDisposed method

rawagner:


rawagner:
retest this

rawagner:
retest this

rawagner:
retest this

rawagner:
retest this

rawagner:
So isDiposed is now impemented. I've also changed some widgets to extend AbstractWidget which resulted in need to change their constructors a bit.  Shell and Menu are the only ones which are not extending AbstractWidget. And I also removed ObjectUtil.InvokeMethod(..."isDisposed")

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/WidgetHandler.java
```diff
@@ -74,6 +74,29 @@ public boolean isEnabled(Widget widget) {
...
+	/**
+	 * Finds out whether specified {@link Widget} is disposed or not.
+	 * 
+	 * @param widget widget to handle
+	 * @return true if widget is disposed, false otherwise
+	 */
+	public boolean isDisposed(Widget widget) {
+		boolean ret = true;
+		Object o = null;
+		try {
```
vpakan: Why use reflection? Why not just directly call widget.isDisposed() method in UIThread?

rawagner:
good point, fixed

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/WidgetHandler.java
```diff
@@ -74,6 +74,22 @@ public boolean isEnabled(Widget widget) {
...
+	 * Finds out whether specified {@link Widget} is disposed or not.
+	 * 
+	 * @param widget widget to handle
+	 * @return true if widget is disposed, false otherwise
+	 */
+	public boolean isDisposed(final Widget widget) {
+		return Display.syncExec(new ResultRunnable<Boolean>() {
+
+			@Override
+			public Boolean run() {
```
vpakan: What about to add check if widget is not null inside the synchExec()?"

rawagner: Display.syncExec method already contains exception handling

vpakan: Yes but question is if we should throw NPE in this case or true. I think better is true because it will make tests more stable not more failing

rawagner: I disagree. Why would you pass a null widget ? I would consider it a mistake/error. I think we should throw exception. We do this for in other methods too so lets keep it consistent

vpakan: That's exactly the point. Sometimes Eclipse dispose some widgets even there is no reason for it. We had couple of issues with it already. So when user is asking if widget is disposed widget is still valid but inside syncExec() variable is set to null. I'm not sure if this can happen but from user point of view if widget is null is for sure disposed

rawagner: I haven't seen Eclipse to set widget to null. Can you reference any issue ?

vpakan: I have no reference to such a problem. For me it's ok to return true when widget is null for me from test point of view if widget is disposed or null it's the same but let's do it your way and don't waste time with any further discussion.

PR #719, opened 2015-11-04
Provide support for debugger testing

apodhrad:
Implement the following operations:
* Add/Remove breakpoints + Breakpoints view
* Start/Finish/Resume debugging
* Check variable names and values in Variable view


apodhrad:
The verification job passed on linux and win. On Mac there was a problem which is not related to this issue. But don't worry I tried it on zamek and it passed as well.

PR #1276, opened 2015-11-05
Keyboard implementation is following CAPSLOCK state

vpakan:
When CAPSLOCK is physically turned on Keyboard implementation types uppercase chars instead of lowercase.


vpakan:
Solution with Java AWT Toolkit doesn't work on Mac will continue with investiagtion

vpakan:
test this

PR #1182, opened 2015-11-09
rev_RedDeer mvn archetype append "reddeer.test" right after provided module/project name

rawagner:
E.g. I want to create a module "org.jboss.reddeer.custom" but I get as a result "org.jboss.reddeer.custom.reddeer.test"


rawagner:
retest this

rawagner:
retest this

PR #1286, opened 2015-11-09
Improve finding labeled widgets

apodhrad:
Currently, the implementation of WithLabelMatcher doesn't find some text fields if the widgets don't create a proper tree. SWTBot solves also such situations.


apodhrad:
This is the first improvement, it the tests pass then I will add the edge case tests

apodhrad:
test this

PR #1148, opened 2015-11-10
jpeterka_Reorganize RedDeer features

rawagner:
RedDeer features are bit messy
* swt.go is part of Eclipse feature should be RCP or could be separate feature
* rcp feature contains to many plugins (Selenium, Gef...)

### DIFF
--- a/features/org.jboss.reddeer.eclipse.feature/feature.xml
+++ b/features/org.jboss.reddeer.eclipse.feature/feature.xml
@@ -91,14 +91,14 @@ Raleigh NC 27606 USA.
          unpack="false"/>
 
    <plugin
-         id="org.jboss.reddeer.selenium"
+         id="org.jboss.reddeer.requirements"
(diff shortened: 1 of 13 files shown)


rawagner:
I didn't add go as a part of eclipse feature because go requires junit.extensions and requirements which are not part of RCP feature

vpakan:
* remove selenium plugin from Eclipse feature
* rename "GO" plugin to "Go"

rawagner:
fixed..hopefully..

jpeterka:
How about selenium feature? Do we want to have it in 1.0 list although it has no API at all (it's entirely empty plugin so far) ?

vpakan:
Selenium is added as separate feature for future use and is not included in any category so it's not visible in install dialog unless user wants to see uncategorized content

jpeterka:
If it's not visible then it's ok for me.

PR #1271, opened 2015-11-11
Method PreferenceDialog.select() takes too long

ljelinkova:
The wait condition takes 10s so pages that do not have the same path as the page name wait very long to be selected. 
```
public void select(String... path) {
		if (path == null) {
			throw new IllegalArgumentException("path can't be null");
		}
		if (path.length == 0) {
			throw new IllegalArgumentException("path can't be empty");
		}
		TreeItem t = new DefaultTreeItem(path);
		t.select();
		
		new WaitUntil(new CLabelWithTextIsAvailable(path[path.length-1]), TimePeriod.NORMAL, false);
	}
```


ljelinkova:
Retest this

ljelinkova:
Retest this

PR #1294, opened 2015-11-11
Update RedDeer product name and update the initial perspective

apodhrad:
Currently the result name is as follows
reddeer-080-linux.gtk.x86_64.tar.gz

It should be reddeer-100-linux.gtk.x86_64.tar.gz


apodhrad:
test this

PR #961, opened 2015-11-12
rawagner_Select menu from toolbar of type HandledContributionItem

rhopp:


rhopp:
test this

rhopp:
test this

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/lookup/MenuLookup.java
```diff
@@ -124,7 +122,47 @@ public ActionContributionItem run() {
...
+								sendHide(item.getMenu(), true);
+								sendShowUI(item.getMenu());
+
+								// drill down
+								currentMenu = item.getMenu();
+								break;
+							} else {
+								
+								// we got a match
+								currentItem = item;
```
rawagner: how about adding a break here ? No need to continue in loop when we already have a match

rhopp: Thanks for noticing. Updated PR is on its way.

rawagner:
retest this

rhopp:
test this

rawagner:
please check the failed test on windows machine, if it has something to do with this PR

PR #1234, opened 2015-11-13
Unify naming convention for exec parameters

apodhrad:
See https://github.com/jboss-reddeer/reddeer/wiki/RedDeer-execution-parameters

### DIFF
--- a/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/properties/RedDeerProperties.java
+++ b/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/properties/RedDeerProperties.java
@@ -13,32 +13,32 @@ import org.jboss.reddeer.common.userprofile.UserProfile;
  */
 public enum RedDeerProperties {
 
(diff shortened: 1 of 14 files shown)


apodhrad:
This PR also requires update the parameters in all jenkins jobs

vpakan:
create issue for updating documentation

apodhrad:
* created at #1305 
* was updated too, see RedDeerLauncherProperties.java
* yes

ljelinkova:
All seems OK except for RedDeerLauncherProperties.getByName. This method was created when we introduced rule that all RD properties have to start with "reddeer" prefix to help to migrate existing launchers. So the method checked if the RD property contains "reddeer" prefix or not. 

The new implementation should be something like this:

{code}
private static RedDeerProperties getByName(String name){
		if (name.startsWith(REDDEER_PARAMETER_PREFIX)){
			return RedDeerProperties.getByName(name);
		} else {
			return RedDeerProperties.getByName(name.replace("reddeer.", "rd."));
		}
	}
{code}

jpeterka:
Seems that all comments were resolved here, let's apply and see then. Pushed to master.

PR #959, opened 2015-12-01
Allow to choose specific columns for details in problems view

mlabuda:


mlabuda:
retest this

mlabuda:
test this

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/ui/problems/ProblemsView.java
```diff
@@ -109,6 +116,82 @@ public ProblemsView() {
...
+	 * Shows specified columns in a tree of problems view. 
+	 * If one of columns is already shown, nothing happens for that one,
+	 * others are shows as expected.
+	 * 
+	 * @param columns columns to show in a tree of problems view
+	 */
+	public void showProblemColumns(Column... columns) {
+		if (columns != null && columns.length > 0) {
+			List<String> problemColumns = getProblemColumns();
+			openConfigureColumnsShell();
```
vpakan: Why not create array of column headers which has to be showed and are currently hidden and than just use new DefaultTable(0).select(<columnsToShowArray>) it's more readable and uses better RedDeer API. And in case no columns are needed to show dialog is not open

The same applies for hideProblemsCountMethod()

PR #1319, opened 2015-12-01
jpeterka_Remove ViewWithTitleIsActive

mlabuda:
This condition probably does not work at all after the re-implementation of View handling (https://github.com/jboss-reddeer/reddeer/commit/3a837b01d3dc955a99a1a4486be3d1269846c03a) and is used only in test that is ignored. 

Probably should be removed.


mlabuda:
test this

PR #1317, opened 2015-12-07
Change RD schema root tag from <testrun> to <reddeer>

rawagner:


rawagner:
already applied to master, closing

PR #1336, opened 2015-12-15
Parametrized doesn't work when constructor is used

rhopp:
Caused by this change: https://github.com/jboss-reddeer/reddeer/pull/1280


jpeterka:
Closed in favor of #1337

PR #1327, opened 2015-12-15
apodhrad_Update logo in RedDeer wizards, UI, etc.

jpeterka:
Also check branding plug-in.


apodhrad:
Could you also update plugins/org.jboss.reddeer.branding/splash.bmp ;-)

apodhrad:
Note that this can be changed later

apodhrad:
ok, let's fix it later

apodhrad:
pushed to master

PR #1341, opened 2015-12-22
rawagner_Incorrect logo in RedDeer recorder

mlabuda:
This happens only in Basic window (when you select RedDeer in the combo box). In the second option it works fine.


mlabuda:
Build passed (compilation).

rawagner:
pushed to master.  Just note that logo is not displayed due to upstream change

PR #1353, opened 2016-01-07
rev_Add copyright to all RedDeer classes

mmalina:


mmalina:
It builds fine for me, so there is nothing to worry about - it's just a comment.

PR #1363, opened 2016-01-21
rawagner_Make TextEditor.getTextEditorPart() method public

vpakan:


rawagner:
Could you change this method from protected to public  https://github.com/jboss-reddeer/reddeer/blob/master/plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/impl/editor/AbstractEditor.java#L292 ?  We would have an option to get eclipse Editor from any editor implementation

PR #1101, opened 2016-02-16
Team import (psf) project import support

mmalina:
Would be nice to have team project import file for RedDeer, check this for an ilustration: https://raw.githubusercontent.com/dgolovin/jbosstools-hibernate/JBIDE-19590/dependencies.psf


mmalina:
There isn't any need for a build for this PR - it just adds a new file.

mmalina:
BTW, here are some links for reference:
http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Ftasks%2Ftasks-cvs-project-set.htm
https://wiki.eclipse.org/Project_set_file

rawagner:
I cant import RedDeer using this psf file. Wizard is expecting .project files but these are not in our sources and we won't add them. Im wondering if there's any possibility to import RedDeer using psf file as maven project ?

mmalina:
After discussion on irc we agreed that we will reject this request, because the only way to make this work would be to have .project everywhere which we don't want.

PR #1370, opened 2016-03-04
Provide reasonable toString() in LogMessage.

rhopp:
This enables for example `logger.error(logView.getErrorMessages())` with reasonable output.


rhopp:
Feel free to merge to 1.1.x branch also.

PR #1372, opened 2016-03-08
Performance improvement of WidgetLookup.activeWidget

scabanovich:
This method is used by AbstractWidget and therefore by all reddeer widgets,
which are created in most integration tests. So, it is one of methods
performance of which is critical.

Previous implementation collected the entire list of matching SWT widgets,
iterating over all widgets in referenced composite and then picked up
an element by the given index. Very often reference composite is not set
so that iteration goes over all widgets of the window.

The suggested implementation does the same iteration but without creating lists
and interrupting the iteration as soon as the requested number of matches is found
(and in most usages the first match is requested).

Otherwise, the logic of the search does not changed. Tests of Reddeer pass.


vpakan:
Can one of the admins verify this patch? (use 'ok to test' comment to run verification job)

rawagner:
retest this

rawagner:
retest this

rawagner:
pushed to master, thanks!

PR #1368, opened 2016-03-08
apodhrad_Evaluate @RunIf after injecting all requirements

mlabuda:
My use case is as follows
```java
@InjectRequirement
private static SwitchYardRequirement switchyardRequirement;
....
@Test
@RunIf(conditionClass = SwitchYardRequirementHasServer.class)
public void createProjectWithServerTest() {
...
}
```
where SwitchYardRequirementHasServer checks whether switchyardRequirement defines a server (using java reflection). Unfortunately, if there is only this one test then the switchyardRequirement is null a the condition returns false.


rawagner:
for 1 .0.x also

apodhrad:
pushed to master and 1.0.x

PR #1384, opened 2016-04-13
rev_RadioButton do not work in Docker Add Connection wizard

jkopriva:
RadioButton in Docker wizard "New Docker Connection" do not correctly, so there is no possibility to use TCP Connection for connecting to a Docker host.

Steps to reproduce:
1. Select perspective "Docker Tooling".
2. In "Docker Explorer" click on "Add Connection".
3. Click on CheckBox "Use custom connection settings:".
4. Click on RadioButton "TCP Connection".
5. Label "Location:" and Button "Browse..." for Unix socket are still accessible, but only "URI:" should be accessible. Validators for URI did not run, so the Finish button is not accessible, if the URI is correct.


vpakan:

rawagner:
test this

PR #1388, opened 2016-04-28
Support publishing Red Deer to JBoss Nexus

rawagner:
To support publishing to JBoss Nexus, you need a distributionManagement section with paths to the snapshot repo:

https://github.com/jbosstools/jbosstools-build/blob/master/parent/pom.xml#L977-L992

### DIFF
--- a/archetype/pom.xml
+++ b/archetype/pom.xml
@@ -67,8 +67,8 @@
 			<uniqueVersion>false</uniqueVersion>
 		</snapshotRepository>
 		<repository>
-			<id>jboss-releases-repository</id>
-			<name>JBoss Releases Repository</name>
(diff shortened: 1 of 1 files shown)


mmalina:
Looking at the JIRA [1] it turns out this is really a problem with nexus itself.
And David Hladky posted a workaround until they find the root cause of this:

there is a temporary workaround - using alternative deployment domain origin-repository.jboss.org (altSnapshotDeploymentRepository switch).
In the meantime we will try to find the problem with the load balancer.

David Hladky

[1] https://issues.jboss.org/browse/JBIDE-22248

nickboldt:
Pull request to work around the problem: https://github.com/jboss-reddeer/reddeer/pull/1389

PR #1389, opened 2016-04-29
issue 1388 / JBIDE-2248 here's a workaround...

nickboldt:
issue 1388 / JBIDE-2248 here's a workaround while the load balancer is busted so you can resume deploying to JBoss Nexus' ORIGIN


nickboldt:
Nexus load balancer appears to be working again so this change is not needed.

PR #1385, opened 2016-05-31
Implement new condition TableHasRow(table, matcher)

jkopriva:
Current implementaion TableHasRows is not enough. Sometimes we need to wait for a specific row.

### DIFF
--- a/plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
+++ b/plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
@@ -7,37 +7,78 @@
  * 
  * Contributors: 
  * Red Hat, Inc. - initial API and implementation 
- ******************************************************************************/ 
+ ******************************************************************************/
 package org.jboss.reddeer.swt.condition;
 
+import java.util.ArrayList;
+import java.util.List;
+
+import org.hamcrest.Matcher;
 import org.jboss.reddeer.common.condition.AbstractWaitCondition;
 import org.jboss.reddeer.swt.api.Table;
+import org.jboss.reddeer.swt.api.TableItem;
 
 /**
  * Condition is met when table has rows.
  * 
- * @author Rastislav Wagner
+ * @author Rastislav Wagner, Josef Kopriva
  */
 public class TableHasRows extends AbstractWaitCondition {
 	private final Table table;
+	private final Matcher<?> matcher;
 
 	/**
-	 * Constructs TableHasRows wait condition.
-	 * Condition is met when table has any rows.
+	 * Constructs TableHasRows wait condition. Condition is met when table has
+	 * at least one row.
 	 * 
-	 * @param table table which should contain rows to the condition be met
+	 * @param table
+	 *            table which should contain rows to the condition be met
 	 */
 	public TableHasRows(Table table) {
+		this(table, null);
+	}
+
+	/**
+	 * Constructs TableHasRow wait condition. Condition is met when table has
+	 * at least one TableItem matched with matcher.
+	 * 
+	 * @param table
+	 *            table which should contain row to the condition be met
+	 * @param matchers
+	 *            matchers - e.g. Regexmatcher, custom matcher...
+	 */
+	public TableHasRows(Table table, Matcher<?> matcher) {
 		this.table = table;
+		this.matcher = matcher != null ? matcher : null;
+	}
+
+	/**
+	 * Search TableItems in table which are matched with matcher.
+	 * 
+	 * @return matchedItems list with matching TableItems
+	 */
+	private List<TableItem> search() {
+		List<TableItem> matchedItems = new ArrayList<TableItem>();
+
+		for (TableItem item : table.getItems()) {
+			if (this.matcher.matches(item)) {
+				matchedItems.add(item);
+			}
+		}
+		return matchedItems;
 	}
 
 	@Override
 	public boolean test() {
-		return table.rowCount() > 0;
+		if (this.table != null && this.matcher == null) {
+			return table.rowCount() > 0;
+		} else {
+			return !search().isEmpty();
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


vpakan:

rawagner:
ok to test

jkopriva:
test this

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
```diff
@@ -10,30 +10,93 @@
...
+	
+	/**
+	 * Constructs TableHasRow wait condition.
+	 * Condition is met when table has specific row and value in specific column matched with matcher.
+	 * 
+	 * @param table table which should contain row to the condition be met
+	 * @param column column in which should be used matchers
+	 * @param matchers matchers - e.g. Regexmatcher, custom matcher...
+	 */
+	public TableHasRows(Table table, Integer column, Matcher<?>... matchers) {
```
apodhrad: This a bit confusing since it is not clear how the matchers are used. Is it enough to meet at least one matcher or all matchers must be met? I suggest to remove the option for specifying more matchers and allow to specify only one matcher. Users can use AndMatchers or OrMatchers (I'm not sure what are the exact names).

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
```diff
@@ -10,30 +10,93 @@
...
+				}
+			}
+		}
+		return !matchedItems.isEmpty();
 	}
 
 	@Override
 	public boolean test() {
-		return table.rowCount() > 0;
+		if(this.table!=null&&this.matchers==null&&this.column==null){
```
apodhrad: Could you add some spaces, please? I recommend to use the auto formatting in Eclipse.

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
```diff
@@ -10,30 +10,93 @@
...
+				int index = 0;
+				while (index < matchers.length && matchers[index].matches(item.getText(i))){
+					index++;
+				}
+				if (index == matchers.length){
+					matchedItems.add(item);
+				}
+			}
+		}
+		return !matchedItems.isEmpty();
```
apodhrad: What about returning the list of matched items?

rawagner:
test this

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
```diff
@@ -7,33 +7,91 @@
...
 public class TableHasRows extends AbstractWaitCondition {
 	private final Table table;
+	private final Matcher<?> matcher;
+	private final Integer column;
 
 	/**
-	 * Constructs TableHasRows wait condition.
-	 * Condition is met when table has any rows.
+	 * Constructs TableHasRows wait condition. Condition is met when table has
+	 * any rows.
```
apodhrad: Any rows (2 and more) or at least one row? I think it's the latter.

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
```diff
@@ -7,33 +7,91 @@
...
+	 * specific row and value in specific column matched with matcher.
+	 * 
+	 * @param table
+	 *            table which should contain row to the condition be met
+	 * @param column
+	 *            column in which should be used matchers
+	 * @param matchers
+	 *            matchers - e.g. Regexmatcher, custom matcher...
+	 */
+	public TableHasRows(Table table, Integer column, Matcher<?> matcher) {
```
apodhrad: We should be more specific of what kind of matcher we require. I think we require only string matchers here, am I right?

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
```diff
@@ -7,33 +7,91 @@
...
+		this.matcher = matcher != null ? matcher : null;
+		this.column = column != null ? column.intValue() : null;
+	}
+
+	private List<TableItem> search(int begin, int end) {
+		List<TableItem> matchedItems = new ArrayList<TableItem>();
+
+		for (TableItem item : table.getItems()) {
+			for (int i = begin; i < end; i++) {
+				if (matcher.matches(item.getText(i))) {
```
rawagner: I suggest using matcher.matches(item) (dont match just item's text, but whole item)

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
rawagner: i think we shouldn't restrain ourselves to only string matchers.  See my comment below

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/TableHasRows.java
apodhrad: In general, this will be a better solution. But in this case it is not necessary. Just out of curiosity, is there any item matcher without matching a text of the the item?

rawagner: you could define specific column to match..not just any

apodhrad: So, we should restrict the matcher to Item.

rawagner:
test this

PR #1379, opened 2016-06-01
Improve exception message in ComboHandler

odockal:
The method setSelection(final Combo combo, final String text) throws only "Nonexisted item in combo was requested". Extend this message by the requested combo text.


rawagner:
retest this

PR #1390, opened 2016-06-14
Screenshots are not created for requirements

mlabuda:
If requirement throw subclass of RuntimeError, screenshot is done. Otherwise not, what causing issues when throwing e.g. AssertionError


rawagner:
test this

rawagner:
test this

PR #1408, opened 2016-07-12
Added new method to TextEditor (fix #1405)

### LINKED ISSUE #1405: New method in TextEditor class that returns line number of searched text
Add implementation of new method for class TextEditor that would take one input String parameter text, searches concrete editor for given text and returns int value of line where the first occurrence was found. If it is not found, -1 is returned instead.

odockal:


plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/TextEditorHandler.java
```diff
@@ -182,6 +182,30 @@ public String run() {
     }
 
     /**
+     * Returns line number of first occurrence of given text.
+     *  
+     * @param editor editor to handle
+     * @param text text to find
+     * @return line number of text found, -1 otherwise
+     */    
+    public int getLineOfText(final ITextEditor editor, final String text) {
```
rawagner: Can you also implement getLineOfTest(editor,text,testIndex) ?

odockal:
New PR with #1410 was created. Closing this one.

PR #1409, opened 2016-07-13
Give some time for "Refreshing server adapter list" job to start

rawagner:


plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/wst/server/ui/wizard/NewServerWizardDialog.java
```diff
@@ -36,14 +41,20 @@ public NewServerWizardDialog() {
...
 		super.finish(timeout);
 	}
 
 	@Override
 	public void cancel() {
 		// workaround due to JBDS-3596
-		new WaitUntil(new JobIsKilled("Refreshing server adapter list"), TimePeriod.LONG, false);
+		Matcher[] jobs = new Matcher[1];
+		jobs[0] = new IsEqual<String>(REFRESHING_JOB_NAME);
+		new WaitUntil(new JobIsRunning(jobs),TimePeriod.NORMAL,false);
```
apodhrad: This logic is already included in the implementation of JobIsKilled. Moreover, if you think that the logic needs a fix then fix it in the JobIsKilled, please.

rawagner: You're right. I didnt notice it. Sometimes i see failures due to refreshing server adapter list. I will have to investigate further

Issue #1403, opened 2016-07-01
Implement Search View class for eclipse class org.eclipse.search2.internal.ui.SearchView


odockal:
Implement reddeer's version of eclipse Search View tab based on org.eclipse.search2.internal.ui.SearchView with appropriate behavior.

rawagner:
pushed to master

PR #1412, opened 2016-07-13
New view class representation of Search View (feature #1403)

odockal:
New PR for #1403.


plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/ui/search/SearchView.java
```diff
@@ -0,0 +1,55 @@
...
+		} catch (CoreLayerException e){
+			return false;
+		}
+	}
+
+	/**
+	 * Returns all search results from Search view as a Tree object with default search option setup
+	 * @return Tree object of search results
+	 */
+	public Tree getSearchResultsAsTree() {
```
rawagner: we are trying to avoid returning swt-layer objects in eclipse layer. I would be better if you implement a new object something like a "SearchResult" which would wrap a single tree item and we would be able to also wrap other methods in the future. 

So this method would return List<SearchResult> and i would call it getSearchResults()

rawagner:
test this

rawagner:
pushed to master

PR #1413, opened 2016-07-14
Add new method into TextEditor (fixes #1405)

### LINKED ISSUE #1405: New method in TextEditor class that returns line number of searched text

odockal:


rawagner:
pushed to master, thanks!
Next time use more descriptive commit message please :)

PR #1414, opened 2016-07-14
rev_Add shell is available condition

rawagner:


plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/impl/shell/AbstractShell.java
```diff
@@ -97,13 +96,9 @@ public boolean isFocused() {
...
 		String text = getText();
 		log.info("Close shell " + text);
-		try {
-			new CancelButton().click();
-		} catch (Exception e) {
-			WidgetHandler.getInstance().notify(SWT.Close, swtShell);
-			ShellHandler.getInstance().closeShell(swtShell);
-		}
-		new WaitWhile(new ShellWithTextIsAvailable(text));
+		//WidgetHandler.getInstance().notify(SWT.Close, swtShell);
```
rawagner: Reasons for update:
- we should not even use Cancel button here, sice method is called close, not cancel
- calling notify and then close is the same thing if you take a look at implementation of close method
- its better/safer to use ShellIsAvailable than ShellWithTextIsAvailable since we already have a shell, not just a text

apodhrad: Please delete the unused code (line 99)

rawagner: fixed

apodhrad:
pushed to master

PR #1420, opened 2016-07-18
Get rid of missing folders 'src' and 'resources' #1419

apodhrad:


apodhrad:
need to link the pr with the existing issue

PR #1419, opened 2016-07-18
rev_Get rid of missing folders 'src' and 'resources'

apodhrad:
If you like the command 'git clean -fd' then you will get errors in Eclipse IDE. The reason is that some projects contain empty src or resources but they are referenced in build.xml.

Possible solutions:
1. Add .gitkeep files
2. Remove the references in build.xml


apodhrad:
I chose the first option

apodhrad:
I killed the verification job since this PR doesn't really need it

PR #1344, opened 2016-07-18
Implement finding an edit part in other edit part

apodhrad:
This will require implementing of WithParent matcher, for an example have a look at Activity implementation ib BPEL tests.

### DIFF
--- a/plugins/org.jboss.reddeer.gef/src/org/jboss/reddeer/gef/matcher/IsEditPartWithParent.java
+++ b/plugins/org.jboss.reddeer.gef/src/org/jboss/reddeer/gef/matcher/IsEditPartWithParent.java
@@ -1,0 +1,57 @@
+package org.jboss.reddeer.gef.matcher;
+
+import org.eclipse.gef.EditPart;
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+
+/**
+ * Returns true if an edit part is a child of a given parent.
+ * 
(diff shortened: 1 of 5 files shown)


plugins/org.jboss.reddeer.gef/src/org/jboss/reddeer/gef/matcher/IsEditPartWithParent.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+
+/**
+ * Returns true if an edit part is a child of a given parent.
+ * 
+ * @author Andrej Podhradsky (apodhrad@redhat.com)
+ *
+ */
+public class IsEditPartWithParent extends BaseMatcher<EditPart> {
```
rawagner: shouldnt this be called EditPartWithParent (without "Is") ? Seems to be more aligned with names of other matches

apodhrad: I dno't see any problem with this naming convention... there are matchers such as WithLabel IsInstanceOf, so why not to use IsEditPartWithParent? For me it is very meaningful.

rawagner: allright :)

rawagner:
pushed to master

PR #1423, opened 2016-07-20
Update RedDeer version to 1.0.3-SNAPSHOT

mmalina:
Command used for upversioning:
mvn org.eclipse.tycho:tycho-versions-plugin:set-version -DnewVersion=1.0.3-SNAPSHOT

In addition, these files were modified manually:
archetype/pom.xml
archetype/src/main/resources/META-INF/maven/archetype-metadata.xml

Ranges were not updated because they are set to [1.0,1.1) which is still correct.


mmalina:
test this

PR #1424, opened 2016-07-22
JBIDE-22757 replace BUILD_ID with...

nickboldt:
JBIDE-22757 replace BUILD_ID with BUILD_TIMESTAMP

### DIFF
--- a/site/pom.xml
+++ b/site/pom.xml
@@ -13,10 +13,8 @@
 	</parent>
 
 	<properties>
-		<!-- Mars convention -->
-		<publishingPath>mars/snapshots/builds/${JOB_NAME}/${BUILD_ID}-B${BUILD_NUMBER}/all/repo/</publishingPath>
-		<!-- Luna convention: TODO JBIDE-18772: remove this when master build only works for Mars -->
-		<publishingPath>builds/staging/${JOB_NAME}/all/repo/</publishingPath>
+		<!-- BUILD_TIMESTAMP is defined in root pom -->
+		<publishingPath>neon/snapshots/builds/${JOB_NAME}/${BUILD_TIMESTAMP}-B${BUILD_NUMBER}/all/repo/</publishingPath>
 		<update.site.name>Red Deer</update.site.name>
 	</properties>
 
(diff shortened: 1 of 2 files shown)


nickboldt:
@mmalina can you take a look at this PR?

site/pom.xml
```diff
@@ -14,7 +14,7 @@
 
 	<properties>
 		<!-- Mars convention -->
-		<publishingPath>mars/snapshots/builds/${JOB_NAME}/${BUILD_ID}-B${BUILD_NUMBER}/all/repo/</publishingPath>
+		<publishingPath>mars/snapshots/builds/${JOB_NAME}/${BUILD_TIMESTAMP}-B${BUILD_NUMBER}/all/repo/</publishingPath>
 		<!-- Luna convention: TODO JBIDE-18772: remove this when master build only works for Mars -->
 		<publishingPath>builds/staging/${JOB_NAME}/all/repo/</publishingPath>
```
mmalina: @nickboldt I'm curious, doesn't this second definition override the first one? And if it does, why isn't anything published here? http://download.jboss.org/jbosstools/builds/staging/reddeer_master/ but this one works:
http://download.jboss.org/jbosstools/mars/snapshots/builds/reddeer_master/
and it does seem to use ${BUILD_TIMESTAMP}-B${BUILD_NUMBER} today.
Would that be because it's overridden in the job? But the job has the neon path set up, not mars:
publishingPath=neon/snapshots/builds/${JOB_NAME}/${BUILD_TIMESTAMP}-B${BUILD_NUMBER}/all/repo/
And, how does maven know about this variable if there is no -D... in the mvn deploy call?

mmalina: Oh, so it does actually work as defined in the jenkins job - the mars url contained an old update site, this is the most current:
http://download.jboss.org/jbosstools/neon/snapshots/builds/reddeer_master/
So could you update the PR to say neon, instead of mars? And remove the luna path.
I still don't understand how mvn deploy gets the variable though. But we can remove that definition in the job once this is merged, right?

nickboldt: Since Luna convention is no longer being used, and since we have the option to override in Jenkins, I've removed the line and updated it to use neon/ instead of mars/.

If you're still building in your 1.0.x branch/job, I'll need to apply this fix there too, but using mars/ instead of neon/ as the folder prefix.

mmalina:
@rawagner can you merge this? thanks!

rawagner:
pushed to master, thanks @mmalina @nickboldt

Issue #1429, opened 2016-07-31
Add new condition class for - LabeledTextIsAvailable


ldimaggi:
Some tests written using Red Deer are experiencing intermittent failures due to timing issues related to new run configurations being created. A solution is to enable the tests to wait for the run configuration "Name:" labeled text to be available. 

This requires the creation of a new class:  org.jboss.reddeer.swt.condition.LabeledTextIsAvailable

rawagner:
Len, you can use WidgetIsFound

`new WidgetIsFound<org.eclipse.swt.widgets.Text>(new ClassMatcher(org.eclipse.swt.widgets.Text.class), new WithLabelMatcher(label))`

PR #1430, opened 2016-07-31
Implemement new class to support waiting for LabeledText

ldimaggi:
See:
https://github.com/jboss-reddeer/reddeer/issues/1429

Some tests written using Red Deer are experiencing intermittent failures due to timing issues related to new run configurations being created. A solution is to enable the tests to wait for the run configuration "Name:" labeled text to be available.

This requires the creation of a new class: org.jboss.reddeer.swt.condition.LabeledTextIsAvailable


rawagner:
See my response in #1429

This condition is not needed because universal solution for all widgets is already implemented - WidgetIsFound

PR #1431, opened 2016-08-01
Replace deprecated ShellWithTextIsActive condition in jface WizardDialog

rawagner:


plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/lookup/ToolItemLookup.java
```diff
@@ -87,7 +87,7 @@ public ToolItem getWorkbenchToolItem(Matcher<String> matcher) {
 		Shell swtShell = ShellLookup.getInstance().getWorkbenchShell();
 		String text = WidgetHandler.getInstance().getText(swtShell);
 		WidgetHandler.getInstance().setFocus(swtShell);
-		new WaitUntil(new ShellWithTextIsActive(text));
+		new WaitUntil(new ShellWithTextIsAvailable(text));
```
mlabuda: On this place waiting for a ShellWithTextIsAvailable is quite useless. Shell is already available and in previous step its getting focused. Original wait condition waited for the shell to be really focused. I would replace those several lines with "new WorkbenchShell()"

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/datatools/ui/wizard/DriverDefinitionWizard.java
```diff
@@ -54,10 +52,9 @@ public void create(DriverDefinition driverDefinition) {
 	 * @see org.jboss.reddeer.jface.wizard.WizardDialog#finish()
 	 */
 	@Override
-	public void finish() {		
-		new DefaultShell(WIZARD_TITLE);
```
mlabuda: I would keep this line. Not sure what user is doing with wizard and maybe it's necessary to keep it here (e.g. to get focus on the shell).

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/lookup/ToolItemLookup.java
rawagner: I changed this one blindly, you're right, waiting for available shell is incorrect. However the whole part of getting shell, setting focus and waiting for it to be active is weird and then I noticed that this method should be deprecated since we dont differentiate between workbench tool item and other tool items and also this method isnt used anywhere anymore. So marked as deprecated :) btw new WorkbenchShell() cannot be used since we are on core layer, workbenchShell is from swt.

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/datatools/ui/wizard/DriverDefinitionWizard.java
rawagner: I use this wizard in hibernate tests and this line is redundant. 
Take a look at super implementation, which doesn't focus shell before clicking on finish and DriveDefinitionWizard is no special case. So I removed it to behave in the same way as other reddeer wizards

PR #1440, opened 2016-08-08
Browser test fails sometimes

rawagner:


rawagner:
So swt browser sometimes returns false when caling some actions (such as back) and that is why test is failing. Im not sure what to do in this situation

mlabuda:
test this

PR #1449, opened 2016-08-15
rev_Deprecate CLabelWithTextIsAvailable in favour of WidgetIsFound

rawagner:


rawagner:
This really does not need verification job, killed

PR #1436, opened 2016-08-19
rev_Add logic for collecting .log output

mlabuda:
- at the beginning of each test clear log
- after test execution collect logs
- store logs in a file
- log processing should be done via Eclipse API, not via UI

### DIFF
--- a/plugins/org.jboss.reddeer.junit.extension/src/org/jboss/reddeer/junit/extension/log/collector/LogCollector.java
+++ b/plugins/org.jboss.reddeer.junit.extension/src/org/jboss/reddeer/junit/extension/log/collector/LogCollector.java
@@ -1,0 +1,131 @@
+/******************************************************************************* 
+ * Copyright (c) 2016 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/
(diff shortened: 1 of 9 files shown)


rawagner:
I tried this locally and looks good to me but Im missing some tests which would verify at least some basic functinality

mlabuda:
I had a class with main method locally and tried it on various examples and it worked. It is a bit difficult to write tests for this because you would have to create a new test class/test suite with test classes which would use RedDeerSuite runner and it would have to be called inside of test method with made up error/some output to log file and then evaluate it on same place. I spoke with @apodhrad  and we agreed its not necessary to have tests for it, that tests for comparator and sorting of prioritized Befores/Afters is enough. Are you ok with this?

plugins/org.jboss.reddeer.junit.extension/src/org/jboss/reddeer/junit/extension/log/collector/LogCollector.java
```diff
@@ -0,0 +1,128 @@
...
+			String line = br.readLine();
+			if (line != null) {
+				bw.write(logDescription + "\n\n");
+				bw.write(line + "\n");
+				while ((line = br.readLine()) != null) {
+					bw.write(line + "\n");
+				}
+				bw.write("\n\n");
+			}
+		} catch (FileNotFoundException e) {
```
rawagner: at least print stacktrace

plugins/org.jboss.reddeer.junit.extension/src/org/jboss/reddeer/junit/extension/log/collector/LogCollector.java
```diff
@@ -0,0 +1,128 @@
...
+		} catch (IOException e1) {
```
rawagner: same here

PR #1443, opened 2016-08-22
rev_Remove org.jboss.reddeer.help plugin and related feature

mlabuda:
It is not necessary anymore. It's almost empty as well and we changed the way documentation is going to be.


mlabuda:
Verification job cancelled - no need to run tests, commit does not affect them. Only compilation is important which has already passed.

site/category.xml
```diff
@@ -17,9 +17,6 @@
 		<description>RedDeer Sources: Contains RedDeer source features.</description>
 	</category-def>
 
-	<feature url="org.jboss.reddeer.help.feature.qualifier" id="org.jboss.reddeer.help.feature" version="0.0.0">
-		<category name="Documentation"/>
```
rawagner: you should also remove category Documentation which is defined above

PR #1471, opened 2016-09-12
TreeViewerHandler can retrieve more matching tree items

mlabuda:
At the moment tree viewer handler just get a single item by its non styled text. If there are several such items in tree, exception is thrown. There should be a method getItems to get all items matching specified path.

### DIFF
--- a/plugins/org.jboss.reddeer.jface/src/org/jboss/reddeer/jface/viewer/handler/TreeViewerHandler.java
+++ b/plugins/org.jboss.reddeer.jface/src/org/jboss/reddeer/jface/viewer/handler/TreeViewerHandler.java
@@ -11,15 +11,16 @@
 package org.jboss.reddeer.jface.viewer.handler;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
 import org.eclipse.swt.custom.StyleRange;
+import org.jboss.reddeer.core.util.Display;
+import org.jboss.reddeer.core.util.ResultRunnable;
 import org.jboss.reddeer.jface.exception.JFaceLayerException;
 import org.jboss.reddeer.swt.api.Tree;
 import org.jboss.reddeer.swt.api.TreeItem;
-import org.jboss.reddeer.core.util.Display;
-import org.jboss.reddeer.core.util.ResultRunnable;
 
 /**
  * Represents handler for TreeViewer widget.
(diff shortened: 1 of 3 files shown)


mlabuda:
I've also cleaned up code. There were 2 methods for retrieving single tree item by single name or full path. I've made universal single method for retrieving single/more tree item(s). Tests are also revamped, tree on which tests are executed is improved. Plus some other changes.

mlabuda:
PR rerun on Mac bcs of timeout, see http://machydra.brq.redhat.com:8080/job/RedDeer_verification_matrix/2634/

plugins/org.jboss.reddeer.jface/src/org/jboss/reddeer/jface/viewer/handler/TreeViewerHandler.java
```diff
@@ -87,74 +56,91 @@ public TreeItem getTreeItem(Tree tree, String name) {
...
 	 */
-	public TreeItem getTreeItem(TreeItem treeItem, String... path) {
-		TreeItem item = treeItem;
-		int index = 0;
+	public List<TreeItem> getTreeItems(TreeItem item, String... path) {
+		return getTreeItems(item.getItems().iterator(), path);
+	}
+	
+	private List<TreeItem> getTreeItems(Iterator<TreeItem> iterator, String... path) {
+		if (path.length == 0 || path == null) {
```
rawagner: null check has to be before length check

plugins/org.jboss.reddeer.jface/src/org/jboss/reddeer/jface/viewer/handler/TreeViewerHandler.java
```diff
@@ -87,74 +56,91 @@ public TreeItem getTreeItem(Tree tree, String name) {
...
+		int index = 0;               
+		Iterator<TreeItem> nestedItemsIterator = iterator;
+		List<TreeItem> resultItems = new ArrayList<TreeItem>();
+	
+		while (index < path.length) {
+			List<TreeItem> nestedItems = new ArrayList<TreeItem>();
+			while (nestedItemsIterator.hasNext()) {
+				TreeItem item = nestedItemsIterator.next();
+				if (getNonStyledText(item).equals(path[index])) {
+					if (!item.isExpanded()) {
```
rawagner: you need to expand item only in else statement
if index==path.length -1 then we dont need to expand tree item. Expanding item is quite lengthy operation

PR #1478, opened 2016-09-21
Widget lookup in Table searches only for TableItems

rhopp:
For example Eclipse Sapphire framework uses Text widgets inside Table. With current implementation there is no way for obtaining DefaultText for this Text inside Table.


rawagner:
have you tried ReferencedComposite ?

rhopp:
Yes... I can put the Table as reference composite, but it still won't search for anything else than TableItems because of line 190 in WidgetResolver ```Widget[] items = ((Table) w).getItems();``` (getItems() returns array of TableItems)

PR #1481, opened 2016-09-22
rev_Improve server state waiting logic

mlabuda:


mlabuda:
@rawagner added also errorMessage for publishing state wait condition.

PR #1484, opened 2016-09-22
Delete project may fail if project is closed

rawagner:


mlabuda:
There is a nice method in Eclipse API. Please implement in direct plugin method isOpen (IProject.isOpen()) and use it in our API.

rawagner:
Fixed

PR #1482, opened 2016-09-23
Make Server tests more stable

mlabuda:
initialization and clean up in before/after instead of before class/ after class etc.


apodhrad:
@mlabuda the PR looks good except the sysout command. Could you have a look at it, please? Once you are done, please add a comment or add the prefix 'rev'.

mlabuda:
This is not ready for a review yet, I saw plenty of jobs failing although locally it passed. I will have to dig deeper.

mlabuda:
Rejected. Locally it looks more stable, but the verification job failed.

PR #1485, opened 2016-09-26
Implement TextEditorHasText

rawagner:


mlabuda:
Would you please rename from TextEditorHasText to TextEditorContainsText? Because condition tells me that editor has specific text but in condition is testing existence of a text in editor texts (as substring).

rawagner:
fixed

mlabuda:
test this

PR #1486, opened 2016-09-26
MouseDoubleClick event is missing count 

rawagner:
Count property for this event is important because org.eclipse.jface.viewers.ColumnViewerEditorActivationEvent does some logic around it


mlabuda:
test this

mlabuda:
test this

PR #1491, opened 2016-10-06
rev_Create a specific category for additional stuff like GEF

mlabuda:
To let user simply uncheck a single category. Otherwise on Eclipse JEE user has to uncheck items in some categories or have installed additional bundles.


mlabuda:
No need to run verification job.

PR #1456, opened 2016-10-11
Allow to work with dirty views

apodhrad:
If a view implements ISaveablePart then it can be dirty (added '*'). Unfortunately, it is not possible to use just the matcher as follows
```
new WithTextMatcher(new RegexMatcher("\\*?" + TITLE)
```


apodhrad:
The verification job has passed

mlabuda:
Please add more tests - set dirty view to be saveable (to let asterisk disappear on save) and then try to get the view. Try to get opened dirty view (open dirty view and then try to get it with instantiating View).

apodhrad:
Added a bit more sophisticated tests, now waiting for the verification job.

apodhrad:
The verification job has passed (except #1489) so the PR is ready to be merged

PR #1411, opened 2016-10-12
rev_Consider using  org.jboss.tools.locus.mockito

apodhrad:
What about using  org.jboss.tools.locus.mockito instead of org.mockito from Orbit?


apodhrad:
Now, it is working with mockito from orbit as well as from locus

PR #1519, opened 2016-11-24
Add Secure Storage Requirement

odockal:
Add new secure storage requirement that switches off master password providers.


rawagner:
@odockal please consider switching off the property using direct API (see org.jboss.reddeer.direct)

odockal:
Going to take a look at it now. Thanks!

odockal:
I did not succeed to find correct plugin id and its key that could switch password providers in Security Storage. I think that concrete providers are generated dynamically.

odockal:
Closing this PR due to insufficient implementation of this area. Will be solved/extended later.

PR #1524, opened 2016-11-26
Some tests that use TaskRespositoriesView.java are seeing intermitten…

ldimaggi:
…t timeouts on Windows only. We should increase the timeout on in the openTask method.

### DIFF
--- a/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/mylyn/tasks/ui/view/TaskRepositoriesView.java
+++ b/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/mylyn/tasks/ui/view/TaskRepositoriesView.java
@@ -90,7 +90,7 @@ public class TaskRepositoriesView extends WorkbenchView {
 		DefaultShell shellThatTakesLongTimeToClose = new DefaultShell("Activate Task");
 		new DefaultText().setText(taskName);
 		new PushButton("OK").click();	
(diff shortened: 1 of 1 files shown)


mlabuda:
Could you also add more tests to RedDeer to catch/see such failures, respectively to don't see it with the fix?

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/mylyn/tasks/ui/view/TaskRepositoriesView.java
```diff
@@ -90,7 +90,7 @@ public void activateTask (String taskName) {
 		DefaultShell shellThatTakesLongTimeToClose = new DefaultShell("Activate Task");
 		new DefaultText().setText(taskName);
 		new PushButton("OK").click();	
-		new WaitWhile(new ShellIsAvailable(shellThatTakesLongTimeToClose), TimePeriod.LONG);
+		new WaitWhile(new ShellIsAvailable(shellThatTakesLongTimeToClose), TimePeriod.VERY_LONG);
```
mlabuda: Please overload this method with additional parameter TimePeriod timeout. So you will be able to set on your own time out in your tests even if it goes higher next time. So you would not have to create a new PR to prolong timeout.

mlabuda:
@ldimaggi any update on this one?

mlabuda:
@rawagner  spotted more problems in this area. This PR does not provide address all of those and is hanging there for ~4 months without changes, thus closing as rejected. Issue #1633 has been created and Rasto will work on that one.

PR #1375, opened 2016-12-05
set junit4 as default in Run Configuration first tab

rawagner:


rawagner:
testPR

rawagner:
@odockal please review

PR #1539, opened 2016-12-09
UI Thread should never wait for state of UI to change, unless it runs…

robstryker:
… event loop

Probably should be sent to master as well.


rawagner:
Can you please fix this in WidgetHandler's getLabel method ? Calling *Handler and *Lookup classes from sync exec is just a bad practise in RedDeer IMHO

rawagner:
also can you open it agains master only ? We dont plan to release 1.2.2 but if we do so, we will cherry-pick this fix from master

PR #1540, opened 2016-12-09
JRE requirements should not have to be changed... check sysprops

robstryker:


rawagner:
also can you open it against master only ? We dont plan to release 1.2.2 but if we do so, we will cherry-pick this fix from master

PR #1551, opened 2016-12-12
UI Thread should never wait for state of UI to change, unless it runs…

robstryker:
… event loop


rawagner:
Thanks for finding this issue. We have added a check here https://github.com/jboss-reddeer/reddeer/blob/master/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/AbstractWait.java#L185 which will notify us if wait is called from UI thread and fixed all such occurrences - including WidgetHandler.getLabel. Fixed in 2.0.0

PR #1556, opened 2016-12-20
Enhance System class to be able to accept ssh keys

rawagner:
connect() method & SystemPasswordPromptDialog class


odockal:
testPR

PR #1559, opened 2016-12-20
Selenium plugin contains typo in package

rawagner:
Selenium plugin (currently empty) has package org.jboss.redderr.bot.selenium. We should fix this typo. It affects also plugin ID in Activator.


mlabuda:
testPR

PR #1575, opened 2017-01-06
Archetype requirements fix

mlabuda:


rawagner:
Archetype is already fixed. You changed the default version project created by archetype which does not have any influence on build. 

archetype is failing because workbench.core plugin is not on update site. But master job on boston jenkins is already running and should be up very soon. See https://github.com/jboss-reddeer/reddeer/commit/b786bfacafc4f144966609ed6e1e782270a4bc8d

rawagner:
Done http://download.jboss.org/jbosstools/neon/snapshots/builds/reddeer_master/2017-01-06_15-44-00-B876/all/repo/plugins/ workbench.core is on update site now.

PR #1580, opened 2017-01-16
Fix sr.getServerNameLabelText(sr.getConfig())

rawagner:
We really do not need to pass the config which serverRequirement already has


rawagner:
testPR

PR #1585, opened 2017-01-20
Add Browse button for rd.config and screenshot dir. Fixes #1465

### LINKED ISSUE #1465: Add a browse button for selecting rd.config
rd_run_config

rawagner:


mlabuda:
Would it be possible to revamp the page under Red Deer tab? When we are moving to 2.0 so the UX could be better. Now we have table and once a user click on a value next to a property, drop down button is shown or browse button is displayed, until then, it is not visible.

mlabuda:
Created followup issue https://github.com/jboss-reddeer/reddeer/issues/1590. Merging and closing this one as resolved.

PR #1578, opened 2017-01-26
Update hierarchy of Servers

mlabuda:


plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/condition/ServerHasPublishState.java
```diff
@@ -11,6 +11,7 @@
 package org.jboss.reddeer.eclipse.condition;
 
 import org.jboss.reddeer.common.condition.AbstractWaitCondition;
+import org.jboss.reddeer.eclipse.wst.server.ui.view.DefaultServer;
```
rawagner: unused import

PR #1591, opened 2017-01-30
Create groups for waiting

mlabuda:
E.g. user want to specify several waitings with a timeout which is common/shared for all of them. E.g. there are 5 waits, first wait for time whole timeout, second wait only for (timeout - timePeriodOfPreviousWaits), third waits only for (timeout - timePeriodOfPreviousWaits) etc. and if (timeout - timePeriodOfPreviousWaits) is lesser or equals to 0, then the wait is performed only once.

### DIFF
--- a/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
+++ b/plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
@@ -1,0 +1,77 @@
+/******************************************************************************* 
+ * Copyright (c) 2017 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/
+package org.jboss.reddeer.common.wait;
+
+import org.jboss.reddeer.common.exception.WaitTimeoutExpiredException;
+
+/**
+ * Group wait consists of various waits in a sequence execution with a common timeout.
+ * Group wait can be used for various processes/actions as a complex wait consisting
+ * of several waitings with 1 shared(decreasing) timeout. If timeout time outs and 
+ * not all waitings passed yet, {@link WaitTimeoutExpiredException} is thrown.
+ *  
+ * @author mlabuda@redhat.com
+ * @since 2.0
+ */
+public class GroupWait {
+
+	private TimePeriod remainingTimeout;
+	
+	/**
+	 * Creates a new group wait with progressive timeout (decreasing) and various wrapped waits.
+	 * 
+	 * @param timeout shared timeout for all waitings
+	 * @param waitings waitings to wait
+	 */
+	public GroupWait(TimePeriod timeout, WaitWrapper... waitings) {
+		remainingTimeout = timeout;
+		if (waitings == null || waitings.length == 0) {
+			throw new IllegalArgumentException("There are no waiting to wait for, pass one or more waitings "
+					+ "as a constructor arguments to start waiting");
+		}
+		for (WaitWrapper waitWrapper: waitings) {
+			long startTimeOfWait = System.currentTimeMillis();
+			if (WaitType.WHILE.equals(waitWrapper.getWaitType())) {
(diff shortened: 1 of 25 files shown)


plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	public GroupWait(TimePeriod timeout, WaitWrapper... waitings) {
+		remainingTimeout = timeout;
+		if (waitings == null || waitings.length == 0) {
+			throw new IllegalArgumentException("There are no waiting to wait for, pass one or more waitings "
+					+ "as a constructor arguments to start waiting");
+		}
+		for (WaitWrapper waitWrapper: waitings) {
+			if (WaitType.WHILE.equals(waitWrapper.getWaitType())) {
+				long startTimeOfWait = System.currentTimeMillis();
```
rawagner: this can be moved just before if condition

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
```diff
@@ -0,0 +1,80 @@
...
+				new WaitWhile(waitWrapper.getWaitCondition(), remainingTimeout, waitWrapper.throwRuntimeException());
+				remainingTimeout = getRemainingTimeoutPeriod(remainingTimeout, startTimeOfWait);
```
rawagner: move outside condition

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
```diff
@@ -0,0 +1,80 @@
...
+			} else {
```
rawagner: else if()

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/WaitProvider.java
```diff
@@ -0,0 +1,26 @@
+package org.jboss.reddeer.common.wait;
```
rawagner: copyright

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
```diff
@@ -0,0 +1,80 @@
...
+				if (WaitType.UNTIL.equals(waitWrapper.getWaitType())) {
+					long startTimeOfWait = System.currentTimeMillis();
```
rawagner: not needed when applied the above suggestion

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
```diff
@@ -0,0 +1,80 @@
...
+					new WaitUntil(waitWrapper.getWaitCondition(), remainingTimeout, waitWrapper.throwRuntimeException());
+					remainingTimeout = getRemainingTimeoutPeriod(remainingTimeout, startTimeOfWait);
+				}
```
rawagner: not needed when applied the above suggestion

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/wait/GroupWait.java
```diff
@@ -0,0 +1,80 @@
...
+	
+	private TimePeriod getRemainingTimeoutPeriod(TimePeriod oldTimeout, long startTimeOfWait) {
+		long diffTime = Math.round((System.currentTimeMillis() - startTimeOfWait) / 1000);
+		long remainingTimeout = oldTimeout.getSeconds() - diffTime;
+		if (remainingTimeout <= 0) {
+			return TimePeriod.NONE;
+		}
+		return TimePeriod.getCustom(remainingTimeout);
+	}
+}
```
odockal: I think, that it would be nice to have GroupWait(Timeout time) constructor for cases where we need to wait for specific time without any special condition...

rawagner: thats what AbstractWait.sleep is for, isnt it ?

odockal: True that.

Issue #1566, opened 2016-12-20
Add check on isDisposed to isVisible method in WidgetLookup


mlabuda:
to make lookup more stable

PR #1603, opened 2017-02-07
Widget disposed check

theJNOVAK:
issue: #1566

### DIFF
--- a/plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/lookup/WidgetLookup.java
+++ b/plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/lookup/WidgetLookup.java
@@ -429,6 +429,10 @@ public class WidgetLookup {
 	 * @return true if widget is visible, false otherwise
 	 */
 	private boolean visible(final Widget w) {
+		if (w.isDisposed()) {
+			return false;
+		}
+		
 		return Display.syncExec(new ResultRunnable<Boolean>() {
 
 			@Override


plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/lookup/WidgetLookup.java
```diff
@@ -429,6 +429,9 @@ public Control run() {
 	 * @return true if widget is visible, false otherwise
 	 */
 	private boolean visible(final Widget w) {
+		if (w.isDisposed())
```
mlabuda: Use braces even if condition is one line

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/WidgetHandler.java
```diff
@@ -104,6 +104,9 @@ public Boolean run() {
 	 * @return true if widget is visible, false otherwise
 	 */
 	public boolean isVisible(Widget widget) {
+		if (widget.isDisposed())
```
mlabuda: Roll back those changes in handler. We do not check whether a widget is disposed in handler methods.

mlabuda:
Next time please use hub to transform an existing issue into a new PR.

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/lookup/WidgetLookup.java
theJNOVAK: edited

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/WidgetHandler.java
theJNOVAK: rolled back

PR #1553, opened 2017-02-08
Remove RCP feature

mlabuda:
@apodhrad do you know any reason to keep RCP feature alive? Me and @rawagner find it useless. If you agree on this, we can remove this feature.


mlabuda:
No need to wait for verification job. Please review.

PR #1626, opened 2017-02-21
Added javadoc generation to install phase

mlabuda:


mlabuda:
No need for verification job. Let it just build and go to verification phase. After merging this one, we will have to edit the job to exclude javadoc:aggregate goal from maven command.

mlabuda:
Unfortunately this easy "solution" does not work. I will have to find another way.

mlabuda:
Well using different phases to bind javadoc did not work, mvn clean install should run in prior to javadoc. There is no post-install phase unfortunately. It is not worth of spending more time on this. Closing as rejected.

Issue #1557, opened 2016-12-13
Fix Graphiti plugin visiblity


mlabuda:
org.jboss.reddeer.graphiti (contains Graphiti exception) and org.jboss.reddeer.graphiti.lookup packages are not exported nor marked as internal. Please fix this.

apodhrad:
Created a new PR at #1634 due to the following error

```
Error creating pull request: Unprocessable Entity (HTTP 422)
fork_collab Fork collab can't be granted by someone without permission
```

PR #1634, opened 2017-03-07
Fix Graphiti plugin visiblity #1557

apodhrad:
See #1557


mlabuda:
Include also org.jboss.reddeer.graphiti  or move exception from that package to org.jboss.reddeer.graphiti.exception and export that one.

apodhrad:
Done

Issue #1555, opened 2016-12-13
Fix package visibility in GEF


mlabuda:
org.jboss.reddeer.gef.comparator is nor marked as an internal package nor exported in MANIFEST. Please fix this.

apodhrad:
Created a new PR at #1635 due to the following error

```
Error creating pull request: Unprocessable Entity (HTTP 422)
fork_collab Fork collab can't be granted by someone without permission
```

PR #1635, opened 2017-03-07
Fix package visibility in GEF #1555

apodhrad:
See #1555


mlabuda:
No need to run verification job.

PR #1507, opened 2017-03-07
Improve the logic for tabbed properties

apodhrad:
Currently, it takes ages to select a tab. This is ok since there are lots of necessary wait conditions. But these conditions are also applied if we try to select already selected tab. I suggest to add some check whether the required tab is already selected and skip the wait conditions.

### DIFF
--- a/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/ui/views/properties/PropertiesView.java
+++ b/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/ui/views/properties/PropertiesView.java
@@ -106,7 +106,9 @@ public class PropertiesView extends WorkbenchView{
 		}catch(Exception ex) {
 			//probably not rendered yet
 		}
-		new WaitUntil(new AnotherTabsRendered(old), TimePeriod.NORMAL, false);
+		if (!old.contains(label)) {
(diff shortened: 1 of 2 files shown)


plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/ui/views/properties/PropertiesView.java
```diff
@@ -106,7 +106,9 @@ public void selectTab(String label) {
 		}catch(Exception ex) {
 			//probably not rendered yet
 		}
-		new WaitUntil(new AnotherTabsRendered(old), TimePeriod.NORMAL, false);
+		if (!old.contains(label)) {
```
mlabuda: use equals instead of contains

apodhrad: 'old' is List and my goal is to check whether the list already contains my label

mlabuda: In that case, it's ok :)

mlabuda:
testPR

apodhrad:
the verification job has failed due to issues unrelated to this PR

PR #1391, opened 2017-03-08
Allow to perform xpath queries in xml files linked to the editor

apodhrad:
This is useful especially in visual editors which work with xml files.


plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/api/EditorFile.java
```diff
@@ -0,0 +1,54 @@
...
+
+	/**
+	 * Evaluates an XPath expression and returns the result.
+	 * 
+	 * @param expression
+	 *            XPath expression
+	 * 
+	 * @return Result of the XPath expression
+	 */
+	String xpath(String expression);
```
mlabuda: what about creating an XMLEditor and doing stuff there? Because in this way, you put getEditorFile to Editor interface so it means we have for any editor following method: editor.getEditorFile().xpath(...), what is not OK, because not every instance of editor can have xpath expression.

plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/impl/editor/internal/BasicEditorFile.java
```diff
@@ -0,0 +1,75 @@
...
+
+	@Override
+	public InputStream getInputStream() {
+		InputStream inputStream = null;
+		try {
+			inputStream = iFile.getContents();
+		} catch (CoreException ce) {
+			throw new WorkbenchLayerException("Cannot get an input stream from the editor file", ce);
+		}
+		return inputStream;
```
mlabuda: wouldn't be better throwing an exception that something went wrong instead of returning null? Also if null iFile is passed in constructor you will get NPE there.

plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/impl/editor/internal/BasicEditorFile.java
```diff
@@ -0,0 +1,75 @@
...
+	 * 
+	 * @param iFile
+	 *            IFile
+	 */
+	public BasicEditorFile(IFile iFile) {
+		this.iFile = iFile;
+	}
+
+	@Override
+	public InputStream getInputStream() {
```
mlabuda: Maybe this method could be private? No need to expose it. And maybe adding a method getResource or getFile which would return iFile to make EditorFile more robust.

apodhrad: I was also thinking of whether to expose input stream or iFile. I decided for input stream since there may be implemnetations which don't use IFile.

plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/api/EditorFile.java
apodhrad: In fact, not only XML editors work with XML files, e.g. SwitchYard editor. So there should be something like XMLBasedEditor which sounds strange and adds more complexity to the logic.

Do you think it is worth for adding a new class? Current implementation just throws an exception that the file cannot be parsed.

mlabuda: Yeah, we should add XMLEditor which would be sibling to TextEditor or its descendant (depends on implementation of shared methods...). We could then enhance our XML editor with more features to have it easier for manipulation - e.g. inserting nodes, inserting elements, setting values, more navigating, searching...

plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/impl/editor/internal/BasicEditorFile.java
mlabuda: And do we need getInputStream public? Maybe making it private/protected and create method to getFile/getResource so users could extend the behavior, if they want to?

plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/api/EditorFile.java
apodhrad: That's exactly the case I don't want. I want the method xpath() for graphical editors based on Gef/Graphiti. Such editors are not XML editors but the file associated with is XML file.

mlabuda: It's not clear from the code that the purpose of this xpath is only for graphical editors. Maybe there should be then such an editor in gef/graphiti plugin. But definitely, it should not be supported by all editors by definition in interface for all editors.

apodhrad: It is not only graphical editors. Note that it could be any editor which works with XML file - text or visual representation of XML file. And it is impossible to detect the underlying file format :(

mlabuda: OK, finally get it ;) I went through code a bit and found out that we have already some xpath evaluation in org.jboss.reddeer.eclipse.wst.xml.ui.tabletree.XMLSourcePage, could you check it? It could be maybe reusable.

I suggest creation of following hierarchy:
We have
- Editor (interface we already have)
- AbstractEditor (abstract implementation of Editor)
- MultiPageEditor (with XML page source and xpath imlementation), TextEditor, ServerEditor, DefaultEditor (Implementations of AbstractEditor)

TODO:
- create interface XMLAssociatedEditor with default implementation of xpath method
- create XMLEditor which extends TextEditor and implements the XMLAssociatedEditor interface - there will go more fancy features over time
- add "implements" of the XMLAssociatedEditor to every editor where it is desired (for existing it is XMLSourcePage, and maybe those graphical editors)

rawagner: @mlabuda  what XMLEditor represents? Is it abstract or not ?

mlabuda: So far we don't have XMLEditor. XMLEditor would be extension for text editor with more XML-oriented PageObject API - insert node, remove node etc.

apodhrad: Implementation of XML editor is out of the scope of this issue

rawagner: TextEditor is based on org.eclipse.ui.texteditor.ITextEditor. It seems that XMLEditor does not exist in eclipse (or name is different ?).

mlabuda: @rawagner  You are right. XMLEditor does not exists in Eclipse, but we could have PageObject API for it ;) 
@apodhrad I dont ask you to implement all fancy methods to XMLEditor, just create a class for further expansion (in follow-ups).

rawagner: I think that in this case defining just interface (with default method) or abstract is enough.

plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/impl/editor/internal/BasicEditorFile.java
apodhrad: If I return IPath then all plugins which use Editor must depend on org.eclipse.core.resources.
Btw what's wrong with the InputStream?

mlabuda: well, it would be similar to Widgets where we return Eclipse SWT.

Nothing wrong is with Input stream. I was pointing that the method to get input stream is used only in another method of the same class. And I am not quite sure whether anybody would use that method outside of the class. If we would provide underlying resource, we provide more power to end users to use it as they want.

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/util/XPathEvaluator.java
```diff
@@ -0,0 +1,152 @@
...
+	 * property 'namespaceAware' to a given value. The input stream is closed
+	 * once the document is built.
+	 * 
+	 * @param inputStream
+	 *            Input stream
+	 * @throws ParserConfigurationException
+	 * @throws IOException
+	 * @throws SAXException
+	 */
+	public XPathEvaluator(InputStream inputStream) throws ParserConfigurationException, SAXException, IOException {
```
mlabuda: In javadoc you declare "setting property 'namespaceAware' but the method signature does not contain such parameter. Basically namespace aware is always set to true.

Could you please add constructor with parameter to set namespace awareness?

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/wst/xml/ui/tabletree/XMLSourcePage.java
```diff
@@ -62,52 +54,12 @@ protected XMLSourcePage(ITextEditor editor) {
 	 * @return the string
 	 */
 	public String evaluateXPath(String xPathExpression){
-		XPath xPath =  XPathFactory.newInstance().newXPath();
-
 		try {
-			Document doc = getDocument();
-			xPath.setNamespaceContext(new UniversalNamespaceResolver(doc));
-			return xPath.compile(xPathExpression).evaluate(doc);
+			XPathEvaluator xPathEvaluator = new XPathEvaluator(new ByteArrayInputStream(getText().getBytes()));
```
mlabuda: You have getAssociatedFile method in Editor. What about overriding the method in this class and call evaluation of xpath in following manner xmlSourcePage.getAssociatedFile.xpath(myExpression)? Because now for some editor you have it done this way. But in this particular its different - ambiguous

apodhrad: There is a difference between getText and getAssociatedFile - what can see in the editor don't need to be in the file, e.g. unsaved state. So, I let the original logic here.

plugins/org.jboss.reddeer.common/src/org/jboss/reddeer/common/util/XPathEvaluator.java
apodhrad: That's the result of merging two various implementations of the same thing ;-) I will fix it asap

apodhrad:
Some other changes have been made, so please take a look once more.

apodhrad:
I have removed the xpath logic from the XMLSourcePage which is now an empty TextEditor. But I let it there so that we can extend it in the future.

apodhrad:
Partially got back the logic in XMLSourcePage + extended the XPathEvaluator by user defined namespaces.

tests/org.jboss.reddeer.common.test/src/org/jboss/reddeer/common/test/util/XPathEvaluatorTest.java
```diff
@@ -0,0 +1,59 @@
+package org.jboss.reddeer.common.test.util;
```
mlabuda: Add license info

plugins/org.jboss.reddeer.workbench/src/org/jboss/reddeer/workbench/impl/editor/internal/BasicEditorFile.java
```diff
@@ -0,0 +1,59 @@
...
+import org.jboss.reddeer.workbench.exception.WorkbenchLayerException;
+
+/**
+ * Basic implementation of EditorFile which represents a file associated with an
+ * editor.
+ * 
+ * @author Andrej Podhradsky (apodhrad@redhat.com)
+ *
+ */
+public class BasicEditorFile implements EditorFile {
```
mlabuda: We left concept of Basic implementation. Use DefaultEditorFile instead, please.

Also consider making it non-internal. Because only public implementation of EditorFile is the XML one. It would be good to have this one as well exposed.

apodhrad: done

tests/org.jboss.reddeer.common.test/src/org/jboss/reddeer/common/test/util/XPathEvaluatorTest.java
apodhrad: done

mlabuda:
testPR

mlabuda:
Please create a doc issue for this PR. Even better, write docs would be ideal ;)

PR #1433, opened 2017-03-08
Add an option to ignore asterisks

apodhrad:
I suggest to extend WithLabelMatcher by an option to ignore asterisks, wdyt?

### DIFF
--- a/plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/matcher/WithLabelMatcher.java
+++ b/plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/matcher/WithLabelMatcher.java
@@ -75,8 +75,15 @@ public class WithLabelMatcher extends BaseMatcher<String> {
 		if ((item instanceof List) || (item instanceof Text) || (item instanceof Button)
 				|| (item instanceof Combo) || (item instanceof CCombo) || (item instanceof Spinner)) {
 			String widgetLabel = WidgetLookup.getInstance().getLabel((Widget)item);
-			if (widgetLabel != null && matcher.matches(widgetLabel)) {
-				return true;
+			if (widgetLabel != null) {
+				// Ignore asterisk and spaces
+				String widgetLabel2 = widgetLabel.trim();
+				if (widgetLabel2.endsWith("*")) {
+					widgetLabel2 = widgetLabel2.substring(0, widgetLabel2.length() - 1).trim();
+				}
+				if (matcher.matches(widgetLabel) || matcher.matches(widgetLabel2)) {
+					return true;
+				}
 			}
 		}
 		return false;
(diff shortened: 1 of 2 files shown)


plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/impl/text/LabeledText.java
```diff
@@ -37,6 +38,6 @@ public LabeledText(String label) {
 	 * @param label the label
 	 */
 	public LabeledText(ReferencedComposite referencedComposite, String label) {
-		super(referencedComposite, 0, new WithLabelMatcher(label));
+		super(referencedComposite, 0, new WithLabelMatcher(new RegexMatcher(label + "\\s*\\**")));
```
mlabuda: I wouldnt use regex matcher inside this in this way. Because if a label contains a special char... it would not be matched properly.

apodhrad: hmm... interesting point... I will also add a test for such case and will see what I can do

apodhrad:
@mlabuda nice catch with the special char! It wasn't working if the label contained '?'. I have changed it a bit so please have a look at it, thx.

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/matcher/WithLabelMatcher.java
```diff
@@ -75,8 +75,15 @@ public boolean matches(Object item) {
 		if ((item instanceof List) || (item instanceof Text) || (item instanceof Button)
 				|| (item instanceof Combo) || (item instanceof CCombo) || (item instanceof Spinner)) {
 			String widgetLabel = WidgetLookup.getInstance().getLabel((Widget)item);
-			if (widgetLabel != null && matcher.matches(widgetLabel)) {
-				return true;
+			if (widgetLabel != null) {
+				// Ignore asterisk and spaces
+				String widgetLabel2 = widgetLabel.trim();
+				if (widgetLabel2.endsWith("*")) {
+					widgetLabel2 = widgetLabel2.substring(0, widgetLabel2.length() - 1).trim();
```
mlabuda: just a syntactic sugar:
if (widgetLabel.trim().endsWith("*")) { matcher.matches(widgetLabel.trim().substring(0, widgetLabel.length() -1).trim()); } else { matcher.matches(widgetLabel); }

apodhrad: Your sugar is not so sweet ;-) Assume the following use case:

Label 'Foo*'

and our code

```
new LabeledText("Foo");
new LabeledText("Foo*");
```

My code work with both constructors but you "sugar" won't work for
```
new LabeledText("Foo*");
```

mlabuda: Yeah, but wasn't it your point to avoid passing asterisks?

apodhrad: My point was to use labels with or without the asterisk. 

There are tests which don't care which fields are required (here we can use labels without asterisk) but there are also tests which take care of which fields are required and which are not (here we need to use the asterisk).

mlabuda: Ok, I understand. In that case it is ok.

Also you should tell developers, which are doing this stuff, to set required field in the right way not by appending the asterisk to label ;)

mlabuda:
testPR

mlabuda:
testPR

PR #1653, opened 2017-03-16
Provide Image to RedDeer Launch Configuration tab

apupier:
- use Eclipse ImageRegistry to save resources and simplify code

image


rawagner:
testPR

rawagner:
pushed to master, thanks!

PR #1637, opened 2017-03-17
ProgressInformationShellIsActive refactor to Available

rawagner:


rawagner:
I decided to remove it completely. User can use ShellWithTextIsAvailable instead.

PR #1670, opened 2017-03-23
Set the package for RedDeer rules as public in RedDeer recorder

apodhrad:
Set the package for RedDeer rules as public in RedDeer recorder so that other plugins (e.g. code generator) can reuse the existing rules.


rawagner:
testPR

PR #1672, opened 2017-03-25
Add Control interface

rawagner:


rawagner:
@odockal some tests failed on ubuntu because of locked screen. Others are ok

PR #1675, opened 2017-03-27
Log view retrieves entries without changing view content. Fixes #1618

### LINKED ISSUE #1618: Shouldnt change content of Log view when getting entries
Now it is done via applying a filter to get only a specific type of entries. Instead of that it would be better to go through entries and find only a specific type of entries (icon next to entry could be a lead).

theJNOVAK:


rawagner:
rebase on master & check javadoc in LogView file because verification job is failing

theJNOVAK:
fixed and rebased

plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/ui/views/log/LogView.java
```diff
@@ -165,27 +119,57 @@ public void restoreLog() {
...
-	
 	private Tree getViewTree(){
 		return new DefaultTree(cTabItem);
 	}
+
+	private List<LogMessage> getMessages(Severity severity){
+		activate();
+
+		ArrayList<LogMessage> messages = new ArrayList<>();
+		for(TreeItem item : new DefaultTree().getItems()){
```
rawagner: LogViewTest.testClearAndRestore failed. Maybe there is a possibility that Tree does not exist when LogView has no messages ?

PR #1680, opened 2017-04-05
ServersView should return actual server type

rawagner:
https://github.com/jboss-reddeer/reddeer/blob/master/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/wst/server/ui/cnf/ServersView2.java#L101 return T


rawagner:
testPR

PR #1676, opened 2017-04-05
merge Shell conditions to ShellIsAvailable

rawagner:


rawagner:
@apodhrad  @mlabuda  unfortunately I cannot merge all Shell conditions to ShellIsAvailable. 
ShellIsAvailable required reddeer shell which comes from swt layer. Other conditions are from core layer. The best I was able to do is to merge ShellWithText and ShellMatchingMatcher to just one condition - ShellMatchingMatcher.

rawagner:
testPR

mlabuda:
We could change it to one condition - ShellIsAvailable, with 3 constructors
public ShellIsAvailable(String title) {...}
public ShellIsAvailable(Matcher matcher) {...}
public ShellIsAvailable(org.eclipse.swt.widgets.Shell shell) {...}

basically, our shell has method to get Eclipse SWT widget so we could call it new ShellIsAvailable(redDeerShell.getSWTWidget());

this way we would have unified constructor with no ambiguity.

@rawagner  @apodhrad   thoughts?

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/condition/ShellMatchingMatcherIsActive.java
```diff
@@ -37,9 +33,9 @@
 	 * 
 	 * @param text title of the shell
 	 */
-	public ShellWithTextIsActive(String text) {
+	public ShellMatchingMatcherIsActive(String text) {
```
apodhrad: I don't like the name, let's user something more general - ShellIsActive

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/condition/ShellMatchingMatcherIsAvailable.java
```diff
@@ -39,13 +38,20 @@ public ShellMatchingMatcherIsAvailable(Matcher<?>... matchers) {
 		AndMatcher am  = new AndMatcher(matchers);
 		this.matcher = am;
 	}
+	
+	/**
+	 * Creates new ShellMatchingMatcherIsAvailable wait condition with specified shell text.
+	 * @param text shell text
+	 */
+	public ShellMatchingMatcherIsAvailable(String text){
```
apodhrad: Again, please use more general name such as ShellIsAvailable. ShellMatchingMatcherIsAvailable says that the parameter must be a matcher (which is not true).

rawagner: We already have ShellIsAvailable.

ShellIsAvailable is on swt.layer and takes RedDeer Shell
ShellMatchingMatcherIsAvailable is on core.layer and accepts String or Matcher. It cannot accept RedDeer Shell since core layer is too low.

So either we keep the matchers as they are now and find more sutiable name for ShellMatchingMatcherIsAvailable or we merge them to one - ShellIsAvailable, move them to core layer and the matcher will no longer accept RedDeer Shell but SWT Shell. WDYT ?

apodhrad: Another option is to move the merged ShellIsAvailable to
```
org.jboss.reddeer.swt.condition
```
Here we could use String, Matcher and RedDeer Shell.

IMHO this is the best solution since the package
```
org.jboss.reddeer.core.condition
```
should not use any condition with SWT dependencies (e.g. SWT Shell).

apodhrad: The above solution would mean moving ShellLookup to reddeer.swt which makes sense for me.

apodhrad: @rawagner @mlabuda let's move all classes from reddeer.core to reddeer.swt since reddeer.core contains only SWT-related classes.

rawagner:
@mlabuda @apodhrad I made some changes to PR.
ShellMatchingMatcherIsAvailable now accepts only matchers
ShellIsAvailable extends ShellMatchingMatcherIsAvailable and accepts matchers,string or shell instance. 
The same applies to ShellIsActive

WDYT ?

rawagner:
testPR

plugins/org.jboss.reddeer.swt/src/org/jboss/reddeer/swt/condition/ShellIsAvailable.java
```diff
@@ -23,20 +25,58 @@ public ShellIsAvailable(Shell shell){
...
+	 * Creates new ShellIsAvailable wait condition with specified matcher.
+	 * @param matchers matcher to match shell title
+	 */
+	public ShellIsAvailable(Matcher<?>... matchers) {
+		super(matchers);
+	}
+	
+	/**
+	 * Creates new ShellIsAvailable wait condition with specified shell text.
+	 * @param text shell text
```
mlabuda: fix param to shellText instead of text, PR jobs is failing.

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/condition/ShellMatchingMatcherIsActive.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class ShellMatchingMatcherIsActive extends AbstractWaitCondition {
+	
+	private AndMatcher matcher;
+
+	/**
+	 * Constructs ShellWithTextIsActive wait condition. Condition is met when
+	 * a shell matching matcher is active.
+	 * 
+	 * @param matcher matcher matching title of the shell
```
mlabuda: please change param matcher to matchers, PR verificaiton job is failing.

rawagner:
@mlabuda  fixed

PR #1678, opened 2017-04-05
Change New, import, export wizard constructors

rawagner:
(shellText, wizardCategory, wizardName)
(shellText, Stringp[ wizardPath)

### DIFF
--- a/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/topmenu/ExportMenuWizard.java
+++ b/plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/topmenu/ExportMenuWizard.java
@@ -10,33 +10,34 @@
  ******************************************************************************/ 
 package org.jboss.reddeer.eclipse.topmenu;
 
-import org.hamcrest.Matcher;
-import org.jboss.reddeer.core.matcher.WithTextMatcher;
+import org.jboss.reddeer.eclipse.ui.dialogs.ExportPage;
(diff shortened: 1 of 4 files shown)


plugins/org.jboss.reddeer.eclipse/src/org/jboss/reddeer/eclipse/topmenu/ExportMenuWizard.java
```diff
@@ -26,25 +26,36 @@
...
+	protected String[] wizardPath;
 	
 	/**
-	 * Constructor set path to the specific item in new wizard dialog.
-	 * @param text dialog text
-	 * @param path to the specific item in new wizard dialog
+	 * Constructs new export Wizard that can be found in ImportExportWizard.
+	 * @param shellText export wizard text when next is clicked
+	 * @param wizardCategory export wizard category
+	 * @param wizardName export wizard name
```
odockal: Would be nice to have examples of wizard category in comment. So the user knows instantly what to use here.

odockal:
Just an idea, but could it not be possible to have one abstract class that would include constructors and methods (that seems to be the same in every class except method open) and than have one abstract method open that would each class implements by its need? Otherwise it seems ok, @odockal approves this one.

rawagner:
sounds good! fixed

odockal:
Thanks for fixing! Just one more detail compilation failures, javadoc requires "&gt;" when using sign of ">"... I believe.. please, check it.

mlabuda:
The problem is with syntax. "\" is not the escape char. To use > and < you should use &gt; and &lt;

rawagner:
@odockal fixed

PR #1650, opened 2017-04-05
Check modal shells before executing any actions

rawagner:


rawagner:
testPR

rawagner:
testPR

plugins/org.jboss.reddeer.core/src/org/jboss/reddeer/core/handler/ControlHandler.java
```diff
@@ -151,5 +158,61 @@ public Composite run() {
...
 		});
 	}
+	
+	/**
+	 * Checks if modal shell that blocks shell in which specified control is located exists 
+	 * and throws exception if it does
+	 * @param control which shell should be checked if it is blocked by some other shell
+	 * @throws CoreLayerException if modal shell that blocks current shell exists
+	 */
+	public void checkModalShells(final org.eclipse.swt.widgets.Control control){
```
apodhrad: I suggest to move the method checkModalShells to AbstractHandler. Classes ControlHandler and WidgetHandler will extend the AbstractHandler.

By doing this we will avoid the following code in widget handlers
```
ControlHandler.getInstance().checkModalShells(getParent(widget));
```

apodhrad:
pushed to master

PR #1709, opened 2017-04-27
Modal shells check may return incorrect value

rawagner:
In this case:

ParentShell -> ChildShell (APPLICATION_MODAL)

when another shell is opened via clicking on button in ChildShell and its parent is again ParentShell then checkModalShells reports that new shell is blocked by ChildShell


rawagner:
testPR

jkopriva:
testPR

PR #1722, opened 2017-05-25
Remove dependency on org.eclipse.m2e.core.ui

apodhrad:
The plugin **org.eclipse.reddeer.swt.test** depends on **org.eclipse.m2e.core.ui**. This dependency is not necessary and we can remove it.


apodhrad:
Hmm... the test ContextMenuTest is fialing after removing that dependency. I will investigate why that happened.

apodhrad:
The test ContextMenuTest.dynamicEnabledAction requires that dependency (it tries to convert a project to a maven project).

apodhrad:
Closed as rejected

PR #1725, opened 2017-05-25
Create an initial target platform for RedDeer development

apodhrad:


apodhrad:
There is an issue related to #1722

apodhrad:
I've added the missing m2e dependency and it should be ok now.

apodhrad:
Added sign-off

rawagner:
please add licence info to the xml file

apodhrad:
Unfortunately, Eclipse IDE removes all comments in that XML file if you open it via IDE :( And we want to open it via IDE and set the TP.

I think it's not necessary to include license info (neither jbosstools does it).

PR #1726, opened 2017-05-25
Revert 'Check modal shells before executing any actions. Fixes #1650'

### LINKED ISSUE #1650: Check modal shells before executing any actions

rawagner:


rawagner:
testPR

rawagner:
Check does not work in this case:

ParentShell -> ChildShell (APPLICATION_MODAL)

when another shell is opened via clicking on button in ChildShell and its parent is again ParentShell then checkModalShells reports that new shell is blocked by ChildShell.

And unfortunately I could not find a way to handle it properly.

PR #1727, opened 2017-05-26
Remove feature junit.extension

apodhrad:
The plugin junit.extension is installed within the feature eclipse. Moreover, the plugin junit.extension depends on the plugina eclipse, workbench, etc...


features/org.eclipse.reddeer.eclipse.feature/feature.xml
```diff
@@ -112,12 +119,5 @@
          install-size="0"
          version="0.0.0"
          unpack="false"/>
-   
-   <plugin
```
rawagner: why did you remove reddeer.go ?

apodhrad:
I've got back the reddeer.go plugin

PR #1723, opened 2017-05-26
Allow to specify BUILD_ALIAS

apodhrad:
Allow to specify BUILD_ALIAS such as
  - CI
  - CR
  - Final

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,6 @@
 	<properties>
 		<tycho-version>0.26.0</tycho-version>
 		<tychoExtrasVersion>${tycho-version}</tychoExtrasVersion>
-		<jbossTychoPluginsVersion>0.26.0</jbossTychoPluginsVersion>
 		<eclipse-target-site>http://download.eclipse.org/releases/neon</eclipse-target-site>
 		<swtbot-update-site>http://download.eclipse.org/technology/swtbot/releases/2.2.1</swtbot-update-site>
 		<eclipse-orbit-update-site>http://download.eclipse.org/tools/orbit/downloads/drops/R20150821153341/repository/</eclipse-orbit-update-site>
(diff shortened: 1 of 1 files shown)


apodhrad:
Now, you can also specify BUILD_TIMESTAMP (by default 'yyyyMMdd-HHmm').So, the version format looks like
{code}
2.0.0.SNAPSHOT-v20170526-1217
{code}

pom.xml
```diff
@@ -124,6 +123,22 @@
...
+				<dependencies>
+					<dependency>
+						<groupId>org.eclipse.tycho.extras</groupId>
+						<artifactId>tycho-buildtimestamp-jgit</artifactId>
+						<version>${tychoExtrasVersion}</version>
+					</dependency>
+				</dependencies>
+				<configuration>
+					<strictBinIncludes>false</strictBinIncludes>
+					<format>'${BUILD_ALIAS}-v'${BUILD_TIMESTAMP}</format>
```
rawagner: why not ${BUILD_ALIAS}.v'${BUILD_TIMESTAMP}. Ie eclipse.core is using dot instead of dash

apodhrad: I used the jbosstools format since eclipse doesn't use the BUILD_ALIAS. Of course, we can discuss which format is best for our purpose.

apodhrad:
Unfortunately, it is not possible to update from
```
2.0.0.SNAPSHOT-vXYZ
```
to
```
2.0.0.Final or 2.0.0.Final-vXYZ 
```

apodhrad:
Changed the default build alias to 'CI' so that a user can update to the final bits.

apodhrad:
Moreover, I think it is ok to have a timestamp for any BUILD_ALIAS.

pom.xml
apodhrad: Moreover, I like the dash ;-)

PR #1713, opened 2017-06-19
RuntimePreferencePage - remove runtime does not have to wait 10s for "Server" shell

rawagner:


rawagner:
@mlabuda  can you take a look ? All test failures are due to oxygen

PR #1743, opened 2017-06-20
Revamped requirements

mlabuda:


genie-reddeer:

rawagner:
testPR

mlabuda:
Wiki page to new requirements https://github.com/eclipse/reddeer/wiki/Requirements-2.0
plus we have to resolve following:
- what to log and format of logging
- nested configurations

mlabuda:
Ok, added support for nested configurations. Amended and force-pushed commit with changes. Created follow-ups #1745 and #1746 which should be addressed ASAP (at least as soon as this PR is merged; before RD 2.0). Job passed green(before amending), yey! 

@rawagner  @apodhrad @odockal please review. PR is ready to review.

mlabuda:
testPR

odockal:
testPR

plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/requirement/matcher/RequirementMatcher.java
```diff
@@ -0,0 +1,92 @@
...
+import org.hamcrest.BaseMatcher;
+
+/**
+ * General Requirement Attribute matcher. Useful for matching requirement's 
+ * attribute to a matching value.
+ * 
+ * @author mlabuda@redhat.com
+ *
+ */
+public abstract class RequirementMatcher extends BaseMatcher<String> {
```
rawagner: can we rename this class to AbstractRequirementMatcher ?

rawagner:
I have created a simple test with @ApacheTomcatServer requirement and added @RequirementRestriction for version >9.0 --> which should not match any of defined servers in json config --> which is OK. However a test should fail (requirement cannot be fulfilled) but it is completely ignored.

rawagner:
pushed to master

PR #1753, opened 2017-06-27
Window should implement ReferencedComposite

rawagner:


rawagner:
testPR

rawagner:
@odockal please take a look.

PR #1749, opened 2017-06-28
Add support for multimatcher requirement config

rawagner:


rawagner:
Supported restrictions are
return type Collection
- only one RequirementMatcher per Requirement
`@RequirementRestriction
public static Collection<RequirementMatcher> restrictions() {
		
		return Arrays.asList(
				new RequirementMatcher(ApacheTomcatServer.class, 
						new RequirementVersionMatcher("family.version", "<9.0")),
				
				new RequirementMatcher(Database.class, new AndMatcher(
						new RequirementRegexMatcher("profileName", "dbProfile1"),
						new RequirementSuperMatcher())));
`
Or return type RequirementMatcher
`@RequirementRestriction
	public static RequirementMatcher restrictions() {
		return new RequirementMatcher(ServerRequirementConfiguration.class, 
				new RequirementVersionMatcher("family.version", "<9.0"));
	}
`

If both are defined an exception is thrown

plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/requirement/matcher/RequirementVersionMatcher.java
```diff
@@ -34,22 +34,17 @@
  * @author mlabuda@redhat.com
  *
  */
-public class RequirementVersionMatcher extends AbstractRequirementMatcher {
+public class RequirementVersionMatcher extends RequirementAttributeMatcher {
+	
+	private String matchingValue;
```
mlabuda: we have matchingValue in both children of RequirementAttributeMatcher. Do you have a reason to not have it in common ancestor together with attributeName?

tests/org.eclipse.reddeer.junit.test/src/org/eclipse/reddeer/junit/test/requirement/matcher/RequirementMatcherTest.java
```diff
@@ -15,148 +15,148 @@
...
 	@Test(expected=RequirementException.class)
 	public void testThrowExceptionOnNonexistingNestedAttribute() {
-		RequirementVersionMatcher matcher = new RequirementVersionMatcher(ComplexConfiguration.class, "simpleConfiguration.what", ">1.0");
+		RequirementMatcher matcher = new RequirementMatcher(ComplexReq.class, new RequirementVersionMatcher("simpleConfiguration.what", ">1.0"));
 		matcher.matches(complexConfig);
 	}
 	
 	@Test
-	public void testRequirementRegexMatcherOnPreciseMatch() {
-		RequirementRegexMatcher matcher = new RequirementRegexMatcher(SimpleConfiguration.class, "name", "SimpleConfig");
```
mlabuda: Why did you removed those 2 tests?

rawagner: the logic for checking the correct class was moved from matchers to https://github.com/rawagner/reddeer/blob/984c47c70257752f4a1c9df4d696f559e804c050/plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/internal/requirement/RequirementHelper.java#L102

plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/requirement/matcher/RequirementVersionMatcher.java
rawagner: String is very restrictive. Maybe i will want my matcher (that extends  RequirementAttributeMatcher) to accept a matcher value, not string value. What would I assign to matchingValue in RequirementAttributeMatcher? 

lets say i will implement
public class MyRequirementMatcher extends RequirementAttributeMatcher {
      MyRequirementMatcher(String attribute, Matcher valueMatcher){
    //what will be assigned to matchingValue in RequirementAttributeMatcher ??
}
}

On the other hand I can say that about attributeName RequirementAttributeMatcher too. So I will move attributeName and atributeValue to RequirementAttributeMatcher and change the type from String to Matcher.

mlabuda: Well, it will always be a string I think, bcs we read values from configuration file (whether it is json now, or yaml later...). Only other thing could be some object in case of nested configurations, but for that we have get nestedAttribute returning String value from config file. So I would keep it as it is at the moment.

tests/org.eclipse.reddeer.junit.test/src/org/eclipse/reddeer/junit/test/requirement/matcher/RequirementMatcherTest.java
mlabuda: Ok,

plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/annotation/AnnotationUtils.java
```diff
@@ -87,32 +89,42 @@
...
+		List<Method> annotatedMethods = Arrays.stream(clazzWithMethod.getDeclaredMethods()).filter(
+				method -> method.isAnnotationPresent(annotationClass)).collect(Collectors.toList());
+		
+		if(annotatedMethods.size() > 1){
+			throw new RedDeerException("More than one method is annotated with "+annotationClass);
+		}
+		if(annotatedMethods.size() == 1){
+			try {
+				boolean methodFound = false;
+				for(Class<?> methodReturnType: methodReturnTypes){
```
rawagner: fixed

plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/requirement/matcher/RequirementVersionMatcher.java
rawagner: You are correct, i cannot change attributeName to Matcher. But i did a few changes and IMHO now its a bit better.
New way of restricting the requirements is:

`new RequirementMatcher(CustomJavaAnnotation.class, "version" , new VersionMatcher("1.8;1.9]"))`

or

`new RequirementMatcher(CustomJavaAnnotation.class, "name" , new RegexMatcher("some_regex"))`


or

`new RequirementMatcher(CustomJavaAnnotation.class, "name" , "some_value")`
which internally translates to IsEqual matcher. This is approach is the same as we do in other places in reddeer :)

notice that I removed AttributeRegexMatcher since simple RegexMatcher can be used. I also renamed AttributeVersionMatcher to VersionMatcher and moved it to common plugin as it can match any string (not tied to Requirements)

plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/requirement/matcher/RequirementMatcher.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * General Requirement Attribute matcher. 
+ * 
+ * @author mlabuda@redhat.com
+ *
+ */
+public class RequirementMatcher extends TypeSafeMatcher<RequirementConfiguration> {
+	
+	private Matcher<?> attributeValue;
```
mlabuda: it could be called attributeMatcher, but no big deal

tests/org.eclipse.reddeer.junit.test/src/org/eclipse/reddeer/junit/test/requirement/matcher/RequirementMatcherTest.java
```diff
@@ -13,150 +13,150 @@
...
-		assertFalse("Name of attribute should not be matched with current regex Complex*", matcher.matches(config));
+		RequirementMatcher matcher = new RequirementMatcher(SimpleReq.class, "name", new RegexMatcher("Complex.*"));
+		assertFalse("Name of attribute should not be matched with current regex Complex*", matcher.matches(simpleConfig));
 	}
 	
 	@Test
 	public void testRequirementVersionMatcherOnPreciseMatch() {
-		RequirementVersionMatcher matcher = new RequirementVersionMatcher(SimpleConfiguration.class, "version", "1.1");
-		assertTrue("Version attribute should be matched to version matcher with value 1.1", matcher.matches(config));
+		RequirementMatcher matcher = new RequirementMatcher(SimpleReq.class, "ersion", new VersionMatcher("1.1"));
```
mlabuda: typo, s/ersion/version/

mlabuda:
2 requested changes, but otherwise great improvements!

rawagner:
both are fixed now!

mlabuda:
Tests passed, there are only oxygen upstream issues.

PR #1689, opened 2017-06-29
add getParentItem to TreeItem

rawagner:


plugins/org.eclipse.reddeer.swt/src/org/eclipse/reddeer/swt/impl/tree/AbstractTreeItem.java
```diff
@@ -124,6 +124,18 @@ public TreeItem getItem(final String text) {
 		return new DefaultTreeItem(treeItemHandler.getItem(swtWidget, text));
 	}
 	
+	/**
+	 * Returns parent tree item
+	 * @return parent tree item
+	 */
+	public TreeItem getParentItem(){
```
rawagner: added & passing on CI

PR #1759, opened 2017-07-03
Repair of failing test ActiveShellExistsTest

odockal:
Just trying to repair one failing test. Might solve [1744](https://github.com/eclipse/reddeer/issues/1744), too.

### DIFF
--- a/tests/org.eclipse.reddeer.core.test/src/org/eclipse/reddeer/core/test/condition/ActiveShellExistsTest.java
+++ b/tests/org.eclipse.reddeer.core.test/src/org/eclipse/reddeer/core/test/condition/ActiveShellExistsTest.java
@@ -40,14 +40,6 @@ public class ActiveShellExistsTest {
 		assertTrue(exists.test());
 		assertEquals(shell.getSWTWidget(), exists.getResult());
 	}
-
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.reddeer.core/src/org/eclipse/reddeer/core/condition/ActiveShellExists.java
```diff
@@ -31,7 +31,7 @@
 	@Override
 	public boolean test() {
 		shell = ShellLookup.getInstance().getCurrentActiveShell();
-		return shell != null;
+		return shell != null ? true : false;
```
rawagner: there is no difference between shell != null and shell != null ? true : false. 

IMHO we have nothing to do here. We simply call SWT API and when it returns null active shell than we do not have a way to find out if really no shell is active

odockal: I thought so. I call test method in assert and get AssertionError: null. I have on idea what is going on here. In any case I am expecting boolean but gets null instead sometimes...

PR #1769, opened 2017-07-18
Add referenced composite to lookup log

rawagner:

### DIFF
--- a/plugins/org.eclipse.reddeer.core/src/org/eclipse/reddeer/core/lookup/WidgetLookup.java
+++ b/plugins/org.eclipse.reddeer.core/src/org/eclipse/reddeer/core/lookup/WidgetLookup.java
@@ -98,7 +98,8 @@ public class WidgetLookup {
 	 */
 	@SuppressWarnings({ "rawtypes","unchecked" })
 	public <T extends Widget> T activeWidget(ReferencedComposite refComposite, Class<T> clazz, int index, TimePeriod timePeriod, Matcher... matchers) {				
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.reddeer.core/src/org/eclipse/reddeer/core/lookup/WidgetLookup.java
```diff
@@ -114,6 +115,17 @@ public static WidgetLookup getInstance() {
 		return (T)found.getResult();
 	}
 	
+	private String createRererencedCompositeDebugMsg(ReferencedComposite refComposite){
+		StringBuilder sb = new StringBuilder();
+
+		if (refComposite == null){
+			sb.append(" no referenced composite specified");
```
mlabuda: Please change appended text to " with no referenced composite specified". So it would result in "Looking up active widget with class type some.package.Widget with no referenced composite specified...."

plugins/org.eclipse.reddeer.core/src/org/eclipse/reddeer/core/lookup/WidgetLookup.java
```diff
@@ -114,6 +115,17 @@ public static WidgetLookup getInstance() {
...
+		} else {
+			sb.append(" referenced composite '"+refComposite.getClass()+"'");
```
mlabuda: Please change appended text to something like " with referenced composite". So it would result in "Looking up active widget with class type some.package.Widget with referenced composite some.other.package.CompositeClass....". Maybe also remove apostrophes, it would look better (class name in whole output is without apostrophes.

rawagner:
both fixed

mlabuda:
testPR

PR #1497, opened 2017-07-20
WizardPages - instead of void return page instance

rawagner:


mlabuda:
Please resolve conflict(s).

PR #1773, opened 2017-07-20
Listen to server changes via IServer API.

rawagner:


rawagner:
seems like this fixed the forever failing server tests

mlabuda:
Yeah, great job!

mlabuda:
I believe, we have more stable Servers now, too :)

PR #1774, opened 2017-07-20
Improve window getEclipseClass method signature

rawagner:

### DIFF
--- a/plugins/org.eclipse.reddeer.jface/src/org/eclipse/reddeer/jface/matcher/WindowMatcher.java
+++ b/plugins/org.eclipse.reddeer.jface/src/org/eclipse/reddeer/jface/matcher/WindowMatcher.java
@@ -12,46 +12,43 @@ package org.eclipse.reddeer.jface.matcher;
 
 import org.eclipse.jface.window.Window;
 import org.eclipse.swt.widgets.Shell;
-import org.hamcrest.BaseMatcher;
 import org.hamcrest.Description;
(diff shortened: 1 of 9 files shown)


plugins/org.eclipse.reddeer.jface/src/org/eclipse/reddeer/jface/matcher/WindowMatcher.java
```diff
@@ -12,46 +12,43 @@
...
+		
 	}
 
 	@Override
-	public void describeTo(Description description) {
-		// TODO Auto-generated method stub
-		
+	protected boolean matchesSafely(Shell item) {
+		try{
+			Object data = WidgetHandler.getInstance().getData((Shell)item);
```
mlabuda: bcs. item is already Shell, it is not necessary to cast Shell on it again. But that's just cosmetics :)

plugins/org.eclipse.reddeer.jface/src/org/eclipse/reddeer/jface/matcher/WindowMatcher.java
```diff
@@ -12,46 +12,43 @@
...
+			if(data != null) {
+				return windowClass.isInstance(data);
+			}
+		} catch (RedDeerException e) {
+			if(((Shell) item).isDisposed()){
```
mlabuda: same here

PR #1785, opened 2017-08-16
Remove the need to implement RequirementConfiguration in all classes

apodhrad:
Assume we want to use a requirement matcher for "fuseServer.host" in the following configuration
```
FuseConfiguration
- camelVersion
- fuseServer
  - host
  - username
  - password
```
At the moment, RedDeer requires the following definitions
```
FuseConfiguration implements RequirementConfiguration
FuseServer implements RequirementConfiguration
```
Do we really want to implement RequirementConfiguration in all inner classes?


plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/util/ReflectionUtil.java
```diff
@@ -0,0 +1,55 @@
+package org.eclipse.reddeer.junit.util;
```
rawagner: copyright

apodhrad: done

PR #1778, opened 2017-08-17
Rename MenuIsEnabled to MenuItemIsEnabled

rawagner:


plugins/org.eclipse.reddeer.core/src/org/eclipse/reddeer/core/handler/MenuItemHandler.java
```diff
@@ -218,4 +218,14 @@ private Event createSelectionEvent(final MenuItem menuItem) {
 		return event;
 	}
 
+	public Menu getMenu(MenuItem swtWidget) {
```
apodhrad: Missing JavaDoc

rawagner:
fixed

PR #1828, opened 2017-09-25
Throw WorkbenchLayerException when save is not enabled (fixes #1816)

### LINKED ISSUE #1816: AbstractEditor.save Method that uses gui to save editor should throws different exception than CoreLayer
Create new exception type SaveNotEnabledException that is derived from CoreLayerException and is thrown if File -> Save operation could not be executed because editor is not dirty. Or add a check that would prevent throwing of CoreLayerException.

odockal:

### DIFF
--- a/plugins/org.eclipse.reddeer.workbench/src/org/eclipse/reddeer/workbench/impl/editor/AbstractEditor.java
+++ b/plugins/org.eclipse.reddeer.workbench/src/org/eclipse/reddeer/workbench/impl/editor/AbstractEditor.java
@@ -32,6 +32,7 @@ import org.eclipse.reddeer.common.wait.AbstractWait;
 import org.eclipse.reddeer.common.wait.TimePeriod;
 import org.eclipse.reddeer.common.wait.WaitUntil;
 import org.eclipse.reddeer.common.wait.WaitWhile;
+import org.eclipse.reddeer.core.exception.CoreLayerException;
 import org.eclipse.reddeer.core.handler.MenuItemHandler;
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.reddeer.workbench/src/org/eclipse/reddeer/workbench/impl/editor/AbstractEditor.java
```diff
@@ -122,7 +123,14 @@ public boolean isDirty() {
 	public void save() {
 		activate();
 		log.info("Save editor");
-		new ShellMenuItem(new WorkbenchShell(), "File", "Save").select();
+		try {
+			new ShellMenuItem(new WorkbenchShell(), "File", "Save").select();
```
rawagner: it would be a lot safer if you call ShellMenuItem.isEnabled -> if false throw ex.

plugins/org.eclipse.reddeer.workbench/src/org/eclipse/reddeer/workbench/impl/editor/AbstractEditor.java
```diff
@@ -122,7 +123,12 @@ public boolean isDirty() {
 	public void save() {
 		activate();
 		log.info("Save editor");
-		new ShellMenuItem(new WorkbenchShell(), "File", "Save").select();
+		ShellMenuItem saveItem = new ShellMenuItem(new WorkbenchShell(), "File", "Save");
+		if (saveItem.isEnabled()) {
```
odockal: @rawagner Very true! Done and thanks!

PR #1848, opened 2017-11-03
Allow passing a reference composite to a context menu

apodhrad:


plugins/org.eclipse.reddeer.swt/src/org/eclipse/reddeer/swt/api/Control.java
```diff
@@ -10,13 +10,15 @@
...
+import org.eclipse.reddeer.core.reference.ReferencedComposite;
+
 /**
  * API for control manipulation
  * @author rawagner
  *
  * @param <T> extends swt Control
  */
-public interface Control<T extends org.eclipse.swt.widgets.Control> extends Widget<T> {
+public interface Control<T extends org.eclipse.swt.widgets.Control> extends Widget<T>, ReferencedComposite {
```
rawagner: Not every Control is not a ReferencedComposite. You can just add constructor to Menu which takes ReferencedComposite.

apodhrad: No, I cannot. Assume the following example
```
new ContextMenu(new DefaultShelll("..."));
```
where the shell is Control and also ReferenceComposite

apodhrad: @rawagner why Control cannot be a reference composite?

apodhrad: Me and @rawagner agreed on adding a method
```
getContextMenu();
```
do editors and controls.

apodhrad:
Rejected

PR #1842, opened 2017-11-03
Add a support for finding a widget according to id

apodhrad:
This was raised on EclipseCon 2017


apodhrad:
Added tests for the most common widgets. Note that it makes no sense to use ID in table/tree items since these widgets are usually constructed during the runtime.

apodhrad:
This PR is ready for review

PR #1850, opened 2017-11-06
Implement getContextMenu in editors

apodhrad:


plugins/org.eclipse.reddeer.workbench/src/org/eclipse/reddeer/workbench/api/Editor.java
```diff
@@ -13,90 +13,112 @@
...
+	 * @return Editor file associated to the editor
+	 */
+	EditorFile getAssociatedFile();
+
+	/**
+	 * Returns a context menu associated to the editor.
+	 * 
+	 * @return Context menu associated to the editor
+	 */
+	Menu getContextMenu();
```
rawagner: we should add this to WorkbenchPart (and make it work with views too)

apodhrad: This can be implemented later without breaking existing API, see #1854

apodhrad: @rawagner so this PR can be merged ;-)

PR #1861, opened 2017-11-29
Fixes regex matcher failures during finding widgets (fixes #1841)

### LINKED ISSUE #1841: Widget lookup with regexp matchers fails to find widgets containing string with special regexp characters
1. Create server adapter with name "CDK 3.2+" (regular name)
2. Try to get the server item in test via ServersView.getServer("CDK 3.2+")
"+" character is the problem here. But I believe that even the dot could cause some troubles. All such characters should be prefixed with "\\" in regexp.

jkopriva:


genie-reddeer:

odockal:
testPR

rawagner:
testPR

odockal:
@rawagner Strange that prcheck job was triggered after your comment, not mine...

jkopriva:
@rawagner thx for merging :)

rawagner:
sure, thx for the fix! :)

PR #1872, opened 2017-12-07
RedDeer 2.0.1 release

rawagner:


tests/org.eclipse.reddeer.workbench.test/META-INF/MANIFEST.MF
```diff
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: RedDeer Workbench Tests
 Bundle-Vendor: Eclipse.org - RedDeer
 Bundle-SymbolicName: org.eclipse.reddeer.workbench.test;singleton:=true
-Bundle-Version: 2.0.1.qualifier
+Bundle-Version: 2.0.1.Final
```
apodhrad: Are you sure we are able to change this?

apodhrad: ah... we did this in each release release, so it should be ok

PR #1878, opened 2018-01-03
Add a logic for code coverage

apodhrad:
Use jacoco which can be then used in SonarQube


apodhrad:
PR is ready to be merged

PR #1888, opened 2018-02-20
Renaming AbstractTable.getSelectetItems

odockal:


odockal:
IMHO, proper way how to deprecate method like this -> method defined in interface, overridden in concrete class that implement the interface, is:
Deprecate the method in interface using javadoc @deprecated tag and using @Deprecated annotation (this will assure that compiler will warn about deprecation of the method), and in child class use @Deprecated annotation on the method so the compiler will notify the user with warning  and do not use javadoc, javadoc from interface/parent class will be used.

Issue #1879, opened 2018-01-18
Create condition BrowserContainsText


jkopriva:
This condition should test if browser contains specific text (e.g. for testing deploying on server and checking, if application is accessible).

This needs to be tested on different platforms.

jkopriva:
Closing - PR is merged.

PR #1898, opened 2018-02-27
Implementation of BrowserContainsText condition (#1879)

jkopriva:

### DIFF
--- a/plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/condition/BrowserContainsText.java
+++ b/plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/condition/BrowserContainsText.java
@@ -1,0 +1,60 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Red Hat, Inc and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.reddeer.eclipse.condition;
+
+import org.apache.commons.lang.StringUtils;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.reddeer.common.condition.AbstractWaitCondition;
+import org.eclipse.reddeer.common.wait.WaitUntil;
+import org.eclipse.reddeer.swt.api.Browser;
+import org.eclipse.reddeer.swt.condition.PageIsLoaded;
+import org.eclipse.reddeer.swt.impl.browser.InternalBrowser;
+
+/**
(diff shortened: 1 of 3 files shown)


genie-reddeer:

genie-reddeer:

odockal:
testPR

plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/condition/BrowserContainsText.java
```diff
@@ -0,0 +1,82 @@
...
+
+	public BrowserContainsText(String text) {
+		this(new InternalBrowser(), null, text, true);
+	}
+
+	@Override
+	public boolean test() {
+		internalBrowser = new InternalBrowser();
+		if (url != null) {
+			internalBrowser.setURL(url);
```
rawagner: seems wrong to set URL everytime a condition is being tested. It should be done only once or it should be left to user. In the second case we should not have a constructor with url parameter.

plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/condition/BrowserContainsText.java
```diff
@@ -0,0 +1,82 @@
...
+		} 
+		if (this.refresh) {
```
rawagner: why refreshing browser in the condition ?

plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/condition/BrowserContainsText.java
```diff
@@ -0,0 +1,82 @@
...
+				pageHTML = (String) internalBrowser.evaluate("return document.documentElement.innerHTML;");
+			}
+			return pageHTML.contains(text);
+		}
+
+	}
+
+	@Override
+	public String description() {
+		return "browser contains text";
```
rawagner: You should also add "text" parameter specified by user

jkopriva:
testPR

jkopriva:
testPR

rawagner:
testPR

plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/condition/BrowserContainsText.java
```diff
@@ -0,0 +1,60 @@
...
+		this.text = text;
+	}
+
+	public BrowserContainsText(String text) {
+		this(new InternalBrowser(), text);
+	}
+
+	@Override
+	public boolean test() {
+		internalBrowser = new InternalBrowser();
```
rawagner: this will always override browser specified by user. Also we should require Browser instead of InternalBrowser

rawagner:
testPR

jkopriva:
@rawagner Thank you for merging and reviews :)

Issue #1894, opened 2018-02-22
Processing close of a shell that was already disposed


odockal:
Sometimes happens that core layer exc. is thrown because there is no shell to handle: https://github.com/eclipse/reddeer/blob/master/plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/wst/server/ui/wizard/ModifyModulesDialog.java#L52. It disappers when entering the if branch.

PR #1901, opened 2018-02-27
Processing closing of a shell that was already disposed

odockal:


odockal:
Linking #1894.

plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/wst/server/ui/wizard/ModifyModulesDialog.java
```diff
@@ -49,9 +49,14 @@ public void finish(TimePeriod timeout) {
 		new FinishButton(this).click();
 		new WaitUntil(new ShellHasChildrenOrIsNotAvailable(getShell()));
 		if(!getShell().isDisposed()){
-			Shell serverShell = new DefaultShell("Server");
-			new OkButton(serverShell).click();
-			new WaitWhile(new ShellIsAvailable(serverShell));
+			try {
+				ShellIsAvailable serverShell = new ShellIsAvailable("Server");
+				new WaitUntil(serverShell);
```
rawagner: how about using WaitUntil(serverShell, TimePeriod.DEFAULT, false)
then if wait's result is not null (meaning shell was found) continue (click ok..)?

PR #1906, opened 2018-03-14
REDDEER-1838 Implement a simple dialog with Ok and Cancel buttons

luvalach:
[REDDEER-1838](https://github.com/eclipse/reddeer/issues/1838) Implement a simple dialog with Ok and Cancel buttons.

Signed-off-by: Lukáš Valach <lvalach@redhat.com>

### DIFF
--- a/plugins/org.eclipse.reddeer.swt/src/org/eclipse/reddeer/swt/impl/shell/OkCancelShell.java
+++ b/plugins/org.eclipse.reddeer.swt/src/org/eclipse/reddeer/swt/impl/shell/OkCancelShell.java
@@ -1,0 +1,97 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Red Hat, Inc and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.reddeer.swt.impl.shell;
+
+import org.eclipse.reddeer.swt.impl.button.CancelButton;
+import org.eclipse.reddeer.swt.impl.button.OkButton;
+import org.eclipse.swt.widgets.Shell;
+import org.hamcrest.Matcher;
+
+/**
+ * Represents a simple shell with "OK" and "Cancel" buttons.
+ * 
+ * @author lvalach
+ *
+ */
+public class OkCancelShell extends DefaultShell {
+
+	/**
+	 * Instantiates a new OkCancelShell.
+	 *
+	 * @param text
+	 *            shell text
+	 */
+	public OkCancelShell(String text) {
+		super(text);
+	}
+
+	/**
+	 * Instantiates a new OkCancelShell.
+	 * 
+	 * @param shell
+	 *            shell with both "OK" and "Cancel" buttons
+	 */
+	public OkCancelShell(Shell shell) {
+		super(shell);
+	}
+
+	/**
+	 * 
+	 * Creates a new OkCancelShell matching specified matcher. First found shell
+	 * with specified matcher is created. Beware, there is no strict (deterministic)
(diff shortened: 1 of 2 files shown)


genie-reddeer:

rawagner:
testPR

genie-reddeer:

plugins/org.eclipse.reddeer.jface/src/org/eclipse/reddeer/jface/dialogs/OkCancelDialog.java
```diff
@@ -0,0 +1,99 @@
...
+import org.eclipse.swt.widgets.Shell;
+import org.hamcrest.Matcher;
+
+/**
+ * Represents a simple dialog with "OK" and "Cancel" buttons.
+ * 
+ * @author lvalach
+ *
+ */
+public class OkCancelDialog extends DefaultShell {
```
rawagner: OkCancelDialog suggests that this represents jface.Dialog class but you are extending DefaultShell (represents swt.Shell) so it should be renamed to OkCancelShell

luvalach: Should I move it also to another package? E.g. org.eclipse.reddeer.swt.impl.shell ?

odockal: Why are you not using TitleAreaDialog as we talked yesterday? Not a good solution?

luvalach: @apodhrad told me that I should extend DefaultShell. If I remember, you also wasn't satisfied with AbstractWindow, you suggested to extend some ***Dialog,  didn't you? It doesn't matter to me, I will do it as you want.

rawagner: @luvalach I did not notice the package :) 
INHO there are two viable options:
1. OkCancelDialog extends AbstractWindow and exists in o.e.r.jface.dialogs package
2. OkCancelShell extends DefaultShell and exists in o.e.r.swt.impl.shell package

@odockal TitleAreaDialog is not a good parent since a simple OK/Cancel dialog does not have to  have Title Area.

odockal: Yes, exactly, TitleAreaDialog was the class. At first you wanted to make it abstract, extending AbstractWindow and that you are right, I did not like the idea, but afterwards, if you can extend TitleAreaDialog which is not abstract, one could use your OKCancelDialog directly for some his concrete use case and keeping eclipse's hiearchy, ie: NewCredentialUserDialog extends org.eclipse.jface.dialogs.TitleAreaDialog which is Dialog with ok and cancel buttons... whyt @apodhrad  @rawagner ?

luvalach:
I renamed it to OkCancelShell, it extends DefaultShell and is located in o.e.r.swt.impl.shell package.

odockal:
testPR

luvalach:
Tests on Oxygen passed, tests on Photon failed but not because of my changes.

plugins/org.eclipse.reddeer.swt/src/org/eclipse/reddeer/swt/impl/shell/OkCancelShell.java
```diff
@@ -0,0 +1,98 @@
...
+public class OkCancelShell extends DefaultShell {
+
+	/**
+	 * Instantiates a new OkCancelShell.
+	 *
+	 * @param text
+	 *            shell text
+	 */
+	public OkCancelShell(String text) {
+		super(ShellLookup.getInstance().getShell(text));
```
rawagner: you can call just super(text), the same applies for constructor with matchers, and also the empty one

luvalach:
Yeah, I did a stupid mistake. Thank you @rawagner, it is fixed now.

Issue #1616, opened 2017-02-10
Add tests for wait conditions


rawagner:

odockal:
Fix merged in #1909.

PR #1909, opened 2018-03-22
Adding tests for waits

odockal:


odockal:
Fixes #1616.

odockal:
@rawagner please note failed test. It might be bug in code instead of failure in test itself, wdyt?

rawagner:
@odockal It is quite a corner case :) Do you dare to add check to GroupWait for nulls ?

odockal:
@rawagner true that... but still, it is there... :)

PR #1912, opened 2018-03-29
RedDeer 2.1.0 release

rawagner:


odockal:
Let see how pr check will end up, but otherwise +1.

PR #1917, opened 2018-04-27
OpenPerspective Req. fails on Photon

rawagner:
Reset Perspective menu item is disabled for a very short time and the default behavior of the requirement is to reset the perspective which fails.


rawagner:
@odockal please take a look

PR #1923, opened 2018-05-16
Remove breakpoints button changed text in photon

rawagner:


rawagner:
@odockal failed tests are not caused by these changes.

plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/debug/ui/views/breakpoints/BreakpointsView.java
```diff
@@ -114,11 +132,10 @@ public void removeAllBreakpoints() {
...
-		new ContextMenuItem("Import Breakpoints...").select();
-		new DefaultShell("Import Breakpoints");
-		new LabeledText("From file:").setText(path);
-		new PushButton("Finish").click();
-		new WaitWhile(new ShellIsAvailable("Import Breakpoints"));
+		new DefaultTree(this).getContextMenu().getItem("Import Breakpoints...").select();
+		WizardDialog importBreakpointsDialog = new WizardDialog("Import Breakpoints");
+		importBreakpointsDialog.isOpen();
+		new LabeledText(importBreakpointsDialog, "From file:").setText(path);
+		importBreakpointsDialog.finish();
```
odockal: Maybe add a check whether finish button is enabled? And if not, throw an exception containing dialog.getMessage() text? One can easily find out what went wrong (file does not exists, was not selected, etc... )

odockal:
@rawagner how about tests? Would you mind write some, uis... ?

rawagner:
@odockal we already have tests org.eclipse.reddeer.eclipse.test.debug.core.DebuggerTest which caught the issue with changed button text. Is there something you are missing ?

odockal:
@rawagner Hm, I missed that test class, alright then, test is at place, great.

plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/debug/ui/views/breakpoints/BreakpointsView.java
rawagner: We will need to create a specific implementation for that dialog because it shows the error message in an unusual place (not at the top, but at the bottom right next to finish/cancel buttons). Lets work in this in another issue.

Issue #1927, opened 2018-05-31
RedDeer example test class missing referenced composite in wizard page


odockal:

odockal:
Also, would be nice to have this class test whether there are no errors... so write a test for it.

PR #1928, opened 2018-05-31
Add missing referenced composite in example test class wizard page

odockal:


odockal:
Fixes #1927

odockal:
@rawagner there is a lot of "Exception during sync execution in UI thread" test failures, among them is one ui test that could have been influenced by the change (but should not be). It is very likely eclipse bug, project explorer tree disposal.

Issue #1936, opened 2018-06-29
Need org.eclipse.reddeer.eclipse.jdt.ui.wizards.NewJavaProjectWizardPageTwo to handle Java projects creation


jeffmaury:
When a test that creates a Java project, the classes `org.eclipse.reddeer.eclipse.jdt.ui.wizards.JavaProjectWizard` and `org.eclipse.reddeer.eclipse.jdt.ui.wizards.NewJavaProjectWizardPageOne`.
It turns out that if the tests is run against Java9/10, the second page will enable the Generate module-info.java checkbox which will leads to a new dialog to be show and causing tests to fail.
This checkbox could be cleared if a `org.eclipse.reddeer.eclipse.jdt.ui.wizards.NewJavaProjectWizardPageTwo` would be provided by RedDeer

PR #1941, opened 2018-07-19
#1936 Upgrade JavaProjectWizard parts

odockal:

### DIFF
--- a/plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/jdt/ui/wizards/JavaProjectWizard.java
+++ b/plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/jdt/ui/wizards/JavaProjectWizard.java
@@ -11,7 +11,16 @@
  *******************************************************************************/
 package org.eclipse.reddeer.eclipse.jdt.ui.wizards;
 
+import org.eclipse.reddeer.common.wait.TimePeriod;
+import org.eclipse.reddeer.common.wait.WaitUntil;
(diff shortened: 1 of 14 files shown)


plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/jdt/ui/wizards/JavaProjectWizard.java
```diff
@@ -25,4 +34,49 @@ public JavaProjectWizard() {
...
+			} else {
+				moduleDialog.dontCreate();
+			}
+			new WaitWhile(moduleShell, TimePeriod.DEFAULT);
+		}
+		
+		new WaitWhile(new WindowIsAvailable(this), timeout);
+		try {
+			new WaitWhile(new JobIsRunning(), timeout);
+		} catch (NoClassDefFoundError e) {
```
rawagner: cannot happen since reddeer.eclipse has a hard dependency on reddeer.workbench :)

odockal:
@rawagner Also removed from WizardDialog.

rawagner:
@odockal WizardDialog is in reddeer.jface which has optional dependency on reddeer.workbench. See https://github.com/eclipse/reddeer/blob/master/plugins/org.eclipse.reddeer.jface/META-INF/MANIFEST.MF#L13 so the try/catch should stay there

odockal:
testPR

odockal:
testPR

odockal:
Tests failures are not connected to this changes, I am going to merge this PR as there were no other required changes from reviewers.

PR #1968, opened 2018-10-30
Adding missing e4 feature into site's categories (fixes #1953)

### LINKED ISSUE #1953: Problems using org.eclipse.reddeer.e4 plugin
I'm currently trying to use the `org.eclipse.reddeer.e4` plugin introduced in https://github.com/eclipse/reddeer/commit/e13032b5dda7a580ae73e1f17cf61c6aa6978c42.
I couldn't find / add it to my dependencies after installing the packages listed in https://github.com/eclipse/reddeer/blob/master/README.md#step-1---install-reddeer-to-eclipse-ide, namely RedDeer Runtime/API and RedDeer UI features. Which package should I additionally install?

I then checked out reddeer from Git and tried to run [WorkbenchPartTest.java](https://github.com/eclipse/reddeer/blob/202a7ddfacac926119b6870fc5df2ee538b3e436/tests/org.eclipse.reddeer.e4.workbench.test/src/org/eclipse/reddeer/e4/workbench/test/part/WorkbenchPartTest.java) but all the tests fail.
The instantiation of [DefaultWorkbenchPart.java](https://github.com/eclipse/reddeer/blob/202a7ddfacac926119b6870fc5df2ee538b3e436/plugins/org.eclipse.reddeer.e4.workbench/src/org/eclipse/reddeer/workbench/impl/part/DefaultWorkbenchPart.java#L23) fails with a NPE because [WorkbenchPartLookup#getWorkbenchPartWithLabel](https://github.com/eclipse/reddeer/blob/202a7ddfacac926119b6870fc5df2ee538b3e436/plugins/org.eclipse.reddeer.e4.workbench.core/src/org/eclipse/reddeer/workbench/core/lookup/WorkbenchPartLookup.java#L64) returns `null`.
To me it looks like those tests can't work since [EPartService#getParts](https://github.com/eclipse/reddeer/blob/202a7ddfacac926119b6870fc5df2ee538b3e436/plugins/org.eclipse.reddeer.e4.workbench.core/src/org/eclipse/reddeer/workbench/core/lookup/WorkbenchPartLookup.java#L66) only returns the part already opened / managed by the `EPartService` and the parts that should get tested don't get opened by the test class.
Can you please tell me what I'm missing?

odockal:


jkopriva:
We can also change 2017 to 2018 in newly added copyright headers.

features/org.eclipse.reddeer.e4.feature/build.properties
```diff
@@ -0,0 +1,13 @@
+###############################################################################
+# Copyright (c) 2017 Red Hat, Inc and others.
```
jkopriva: @odockal please update newly added copyright to 2018, thank you :)

odockal:
Last commit should have contain fix for archetypes that was failing recently... ide workbench is no starting -> solution is to add org.eclipse.equinox.ds into target-platform-configuration into pom.xml of archetype project. One cannot use 1.3.0-SNAPSHOT tycho as it does not exist in used sonatype repositories.

odockal:
@jkopriva PR check is back to unstable (not erroneous). Archetype can be built now. Can we merge if no other objection will appear? Thanks!

PR #1988, opened 2019-01-18
Issue-1975 Fixes for JDK11 compilation failures

jkopriva:


odockal:
testPR

odockal:
@jkopriva  I think we are good to go! Thanks.

Issue #1990, opened 2019-01-28
Returning collection of requirement matchers is not properly processed when the same requirement class used


odockal:
[Check](https://github.com/eclipse/reddeer/blob/37f27affd709a9622d84e8d29d047dc78a0de1b4/plugins/org.eclipse.reddeer.junit/src/org/eclipse/reddeer/junit/internal/configuration/SuiteConfiguration.java#L135) method is evaluating collection of req. matchers in a wrong way. It throws and exception when there is more than one req. matcher with the same req. class, but should be thrown when there is the same req. class used with the same attribute.

Issue #1991, opened 2019-01-28
Requirement matchers returned in collection of the same class are evaluated as in OR relation


odockal:
More than one requirement matcher of the same class returned in collection in RequirementRestriction annotated method are evaluated as in OR relation. Thus, there is more configurations returned than it should from cartesian product of requirement combination. In this [line], code is iterating over each passed requirement, but it should be grouped by configuration class of the requirement and tested whether all attributes of requirement matcher fulfills the condition. 

[line]: https://github.com/rawagner/reddeer/commit/aec2742fde7b157296b90540ad3b8749bd1863d0#diff-9ac0d54517547dc1b47e6c93512260d7R101

PR #1992, opened 2019-01-29
Requirement restriction editing, fixes issue #1990

odockal:


jkopriva:
@odockal please review failing tests, thank you!

odockal:
I am going to ask you, whether you would be that nice and check the tests themselves, whether they are ok. In case we agree, test will be repaired with PR fixing #1991.

odockal:
the main question here is, what should static method annotated with RequirementRestriction returning a collection of requirement matchers do, when there are more than one requirement matcher with the same configuration class but different attributes that the matcher aims, and when they are matching the same attribute? (Second case is quite obvious, as we do not want to have multiple req matchers (with the same config class) testing same attributes).

PR #1999, opened 2019-02-10
Change relation between requirement matchers from OR to AND (fixes #1991)

### LINKED ISSUE #1991: Requirement matchers returned in collection of the same class are evaluated as in OR relation

odockal:
- contains fixes for tests failures introduced in PR for #1990


odockal:
Fixing #1991.

odockal:
testPR

odockal:
@jkopriva please, take a look.

odockal:
Closing in favor of #2000.

PR #2000, opened 2019-02-12
Change relation between requirement matchers from OR to AND

odockal:


odockal:
Fixes #1991

PR #2009, opened 2019-04-16
Align RD 2.6.0 to Eclipse 2019-06

odockal:
- New File dialog changed to Create New File
   - bump up version to 2.6.0


odockal:
@jkopriva Only failed test is old friend of ours. (Unstable manipulateShell). I think we are good to go.

